PR #26, opened 2014-10-20
WIP: new timing loops

japaric:
This PR is still missing some documentation, but the new timing loops are already useable.

I found some surprising/interesting results when testing the new timing loops. I'm currently analyzing the results, I'll post my notes here when I'm done.

cc @kmcallister


japaric:
rebase in #62

Issue #55, opened 2016-01-20
restarting development and criterion's future


japaric:
I'd like to get criterion back into a usable state, and I'd like to improve the
user experience around it.

Here's my plan so far, input is appreciated and PR are also welcomed.
### Immediate actions: criterion 0.1.0

Basically, release criterion with its current public API, refactor its
internals, further improvements shouldn't break the API.
- [x] land new timing loops #26 
- [x] make compilable on current nightly
- [x] refactor: use the new Instant/Duration APIs (when stable?)
- [x] fix travis
- [x] CI: test on mac. see #70 
- [x] CI: test on windows. see #70 
- [x] documentation: explain output/plots, add examples, how to use in a cargo
  project
- [x] fix Appyevor
- [x] release dependencies
- [x] start making releases.
### Short term: criterion 0.1.x
- [x] refactor: reduce usage of unstable features.
  - Throw away or put simd behind a cargo feature, it's unstable and only works
    with x86_64
  - criterion won't work on stable anytime soon because it depends on
    test::black_box
- [x] colorize output messages
- [x] developer documentation: explain the math
- [x] refactor: proper error handling in internals. bubble up errors and report them instead of panicking/unwrapping
- [x] clippy
- [x] rustfmt
### Future: criterion 0.2

If we are allowed breaking changes, how can we improve the user experience? Some
ideas:
- [x] simplify the public API
  - #26 adds several methods, perhaps all the similarly named methods can be
    replaced with a function that takes enums or some builder pattern.
- [ ] expose errors to user instead of panicking/exiting
- [ ] expose more internals, e.g. return an intermediate struct with the benchmark results
  instead of directly writing the results to disk.
- [ ] move away from gnuplot, produce a web report with interactive plots
- [x] better integration with cargo, output files to target directory, plots
  could live in target/doc

If you used criterion in the past, I'd like to hear about your experience.
- In general, What worked well? And, What didn't work for you?
- How helpful/clear/confusing were the generated plots/the output message?
- Was the performance regression detection reliable, or did you get false
  positives?
- What functionality do you think is missing?
- Any suggestion to improve the user experience?

japaric:
@faern will, hopefully this week, send us some PR to get us close to compiling on a recent nightly. :+1:

japaric:
Next actions:

- publish criterion dependencies
- write great docs
- publish criterion 0.1.0

[...]

PR #56, opened 2016-01-25
Updating criterion.rs to work with rust nightly 2015-11-27

faern:
In response to issue #55. I'm not sure every fix does exactly what it's supposed to. I mostly did my best to make it build.

This PR changes some of the dependencies to other forks of the dependencies that are more updated. These changes should of course be merged to the upstream dependencies and this crate be pointed back to them, but I started here and we can work on it from within this PR maybe?

### DIFF
--- a/src/routine.rs
+++ b/src/routine.rs
@@ -1,5 +1,3 @@
-use std::iter;
-
 use time;
 
 use format;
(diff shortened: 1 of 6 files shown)


src/routine.rs
```diff
@@ -24,17 +22,20 @@ pub trait Routine {
...
         // Solve: [d + 2*d + 3*d + ... + n*d] * met = m_ns
-        let d = (2. * m_ns as f64 / met / (n * (n + 1)) as f64).ceil() as u64;
+        let total_runs = n * (n + 1) / 2;
+        let d = (m_ns as f64 / met / total_runs as f64).ceil() as u64;
 
-        let m_iters = iter::iterate(d, |a| a + d).take(n).collect::<Vec<u64>>();
+        let m_iters = (1..n as u64).map(|a| a*d).collect::<Vec<u64>>();
 
-        let m_ns = m_iters.iter().map(|&x| x).sum::<u64>() as f64 * met;
+        let m_ns = total_runs as f64 * met;
```
japaric: This should be `let m_ns = total_runs as f64 * d as f64 * met`

japaric:
I've looked over these changes and the forked dependencies, and all the updates seem to preserve the behavior except for the one that I pointed out here. I'd be in favor of merging this PR as a first step forward if @laumann is okay with us using his commits.

laumann:
@japaric @faern No problem with me :smile: Glad that you can use it...

japaric:
@laumann Great, thanks :smile:

PR #57, opened 2016-02-02
Rustup: Update dependencies and make structs public

faern:
I pointed some dependencies back to your repos since they should work now.

Do you agree with making the structs public? I'm mostly guessing to make stuff build again. The sound like they could be public. Note that I didn't provide the best docs yet. Please come with suggestions on what to write or say that it's OK to leave as is. Linter options force us to provide docs.

@japaric

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,14 +14,14 @@ git = "https://github.com/rust-lang/log"
 git = "https://github.com/rust-lang/rustc-serialize"
 
 [dependencies.simplot]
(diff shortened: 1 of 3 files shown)


japaric:
> Do you agree with making the structs public?

No, they are an implementation detail and shouldn't be in the public API. I'm okay with merging this as it is though, we can work on making them private again before 0.1.

japaric:
Thanks!

PR #58, opened 2016-02-06
compile with nightly rustc

Alex-Diez:


japaric:
Thanks!

japaric:
BTW, walkdir should have been listed as a dev-dependency but I fixed that in a follow-up commit 7e64bbd89e252d5b53b63e98445a2b19f72a7368

PR #62, opened 2016-02-09
new timing loops take 2

japaric:
rebase of #26 

I've kept the method names but we probably want to change them to something more meaningful before landing this.

I've also changed the documentation, hopefully is more useful this time.

cc @faern

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -30,10 +30,10 @@ mod report;
 mod routine;
 
 use std::default::Default;
-use std::fmt;
 use std::iter::IntoIterator;
 use std::process::Command;
 use std::time::{Duration, Instant};
+use std::{fmt, mem};
 
 use rustc_serialize::json;
 use std::fs::File;
@@ -42,7 +42,10 @@ use std::path::Path;
 
 use estimate::{Distributions, Estimates};
 
-/// Helper struct to build functions that follow the setup - bench - teardown pattern
+/// Helper struct to time routines
+///
+/// This struct provides different "timing loops" as methods. Each timing loop provides a different
+/// way to time a routine and each has advantages and disadvantages.
 #[derive(Clone, Copy)]
 pub struct Bencher {
     iters: u64,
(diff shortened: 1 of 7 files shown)


src/lib.rs
```diff
@@ -60,6 +88,166 @@ impl Bencher {
...
+    pub fn iter_with_setup<I, O, S, R>(&mut self, mut setup: S, mut routine: R)
+        where S: FnMut() -> I,
+              R: FnMut(I) -> O
+    {
+        self.elapsed = Duration::from_secs(0);
+        for _ in 0..self.iters {
+            let input = setup();
+
+            let start = Instant::now();
+            let output = test::black_box(routine(input));
```
faern: Would it make sense to add `black_box` around the input as well? I'm not sure about this, just want to prevent rustc to optimize the call `routine(input)`

src/lib.rs
```diff
@@ -60,6 +88,166 @@ impl Bencher {
...
+    /// Times a `routine` that needs to consume its input by first creating a pool of inputs.
+    ///
+    /// This function is handy for benchmarking destructors.
+    ///
+    /// WARNING This requires `iters * mem::size_of::<I>()` of memory, and `iters` is not under the
+    /// control of the caller.
+    ///
+    /// # Timing loop
+    ///
+    /// ```
```
faern: Must be `ignore` for tests to run

src/lib.rs
```diff
@@ -60,6 +88,166 @@ impl Bencher {
...
+    /// start = Instant::now();
+    /// for input in inputs {
+    ///     routine(input);
+    /// }
+    /// elapsed = start.elapsed();
+    /// ```
+    ///
+    /// # Timing model
+    ///
+    /// ```
```
faern: Set to `text` maybe? Tests won't run

japaric:
@faern Any suggestions about changing the methods' names?

faern:
@japaric You mean for the `iter`-methods? No I think they are quite descriptive.

PR #64, opened 2016-02-23
New iter function and benchmarking over multiple functions

faern:
I hope I remembered to add everything I needed here.

I both added an iter function and the possibility to benchmark multiple functions. Do you think both should be included?

Example of where I use this: https://github.com/faern/forkjoin-benchmarking/blob/master/src/main.rs#L108

@japaric

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -42,15 +42,19 @@ use std::path::Path;
 
 use estimate::{Distributions, Estimates};
 
-/// TODO
+/// Representing a function to benchmark together with a name of that function.
+/// Used together with `bench_compare_implementations` to represent one out of multiple functions
+/// under benchmark.
 pub struct Fun<I: fmt::Display> {
     n: String,
     f: Box<FnMut(&mut Bencher, &I)>,
 }
 
-impl<I: fmt::Display> Fun<I> {
-    /// TODO
-    pub fn new<F: FnMut(&mut Bencher, &I) + 'static>(name: &str, f: F) -> Fun<I> {
+impl<I> Fun<I> where I: fmt::Display {
+    /// Create a new `Fun` given a name and a closure
+    pub fn new<F>(name: &str, f: F) -> Fun<I>
+        where F: FnMut(&mut Bencher, &I) + 'static
+    {
         Fun {
             n: name.to_string(),
             f: Box::new(f),
@@ -493,13 +497,35 @@ impl Criterion {
         self
     }
 
-    /// TODO
+    /// Benchmarks multiple functions
+    ///
+    /// All functions get the same input and are compared with the other implementations.
+    /// Works similar to `bench`, but with multiple functions.
+    ///
+    /// ``` ignore-test
+    /// fn bench_seq_fib(b: &mut Bencher, i: &u32) {
+    ///     b.iter(|| {
+    ///         test::black_box(seq_fib(test::black_box(i)));
+    ///     });
+    /// }
+    ///
+    /// fn bench_par_fib(b: &mut Bencher, i: &u32) {
+    ///     b.iter(|| {
+    ///         test::black_box(par_fib(test::black_box(i)));
+    ///     });
+    /// }
+    ///
+    /// let sequential_fib = Fun::new("Sequential", bench_seq_fib);
+    /// let parallel_fib = Fun::new("Parallel", bench_par_fib);
+    /// let funs = vec![sequential_fib, parallel_fib];
+    ///
+    /// Criterion::default().bench_compare_implementations("Fibonacci", funs, &14);
+    /// ```
     pub fn bench_compare_implementations<I>(&mut self,
         id: &str,
         funs: Vec<Fun<I>>,
         input: &I) -> &mut Criterion
-        where
-        I: fmt::Display
+        where I: fmt::Display
     {
         analysis::functions(id, funs, input, self);
         self
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -42,6 +42,22 @@ use std::path::Path;
 
 use estimate::{Distributions, Estimates};
 
+/// TODO
+pub struct Fun<I: fmt::Display> {
+    n: String,
+    f: Box<FnMut(&mut Bencher, &I)>,
+}
+
+impl<I: fmt::Display> Fun<I> {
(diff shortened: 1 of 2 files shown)


faern:
Documentations should probably be added. But I wanted some feedback on the code first.

japaric:
I'm not sold on `iter_with_setup_and_verify`. IMO verification of a function/routine correctness doesn't belong in a benchmarking library, and it's best left to a testing library like [quickcheck].

[quickcheck]: https://github.com/BurntSushi/quickcheck

The `bench_compare_implementations` LGTM. But I have yet to check what the summary plots look like, if you have one at hand that you can share here it would great.

re: docs. They should mention when to use this `compare` method and an example would be nice.

nit: I prefer to move trait bounds to `where` clauses whenever possible.

faern:
Examples of plots over multiple functions: https://imgur.com/7ZQjqD3 and https://imgur.com/uDMg2GM. The names I had on the functions here were a little bit boring. What I compared was a sequential algorithm vs the same algorithm running on a thread pool with different amounts of threads in them (T1 to T32).

I were almost expecting you to not like `iter_with_setup_and_verify`, I can remove it for now since it's not very related to this PR anyway. I just found it a little bit frustrating to not being able to get the result out of the benches in any way when I used criterion.rs. When I was doing development I just ran the benchmarks to both verify correctness and speed.

I'll fill in some docs and have a look at the trait bounds.

japaric:
I left a question, depending on your answer we could update the example in the doc comment.

Other than that this looks ready to go.

It just occurred to me that we may want a method that tests multiple functions over multiple inputs. This could be used to bench a sequential vs a parallel implementation of some routine under increasing input size. The seq impl will probably be faster for small inputs (because of the thread spawning overhead) and the parallel impl will be faster for bigger inputs. This bench would tell you at what size it makes sense to switch from the seq impl to the parallel one.

faern:
I had that thought about testing multiple functions over multiple inputs back in last spring when I used criterion.rs more. But back then I didn't have neither the time nor the knowledge to implement that by myself.
That would be awesome. But I think we can take that after 1.0.

src/lib.rs
```diff
@@ -477,6 +497,40 @@ impl Criterion {
...
 
+    /// Benchmarks multiple functions
+    ///
+    /// All functions get the same input and are compared with the other implementations.
+    /// Works similar to `bench`, but with multiple functions.
+    ///
+    /// ``` ignore-test
+    /// fn bench_seq_fib(b: &mut Bencher, i: &u32) {
+    ///     b.iter(|| {
+    ///         seq_fib(test::black_box(i));
```
japaric: I still think that's unfortunate that we have to use `black_box` at all :P.

japaric:
> That would be awesome. But I think we can take that after 1.0.

Ah, sure after 0.1.0. I don't think is that hard to implement. What worries me is the name :P. `bench_compare_implementations_with_multiple_inputs`? I'm sketching a builder pattern for `Criterion` that would, in theory, remove the need for several `bench` methods but that's material for 0.2.0 or we could add a `v2::Criterion` to the 0.1.0 branch.

src/lib.rs
faern: Ok. I don't get it. Do you oppose the usage of `black_box` in general or just in this case? I'm not sure if we need it there or not. I'm mostly tired of benchmarks getting optimized away so I sprinkle them like stardust on all my code when I benchmark.

faern:
These names get out of hand. There must be a different way to think about them in general. Something that would be shorter, but not optimal would be this set of names
* `bench_function` (current `bench`)
* `bench_functions`
* `bench_functions_over_inputs`

src/lib.rs
japaric: > Do you oppose the usage of black_box in general or just in this case?

IMO, ideally, the _users_ should never need to use `black_box` in their benchmarks. It's cumbersome, they need import the `test` crate and add a feature gate. I'm OK with sprinkling `black_box` in all of criterion's internals because that's not user facing.

But if your experience says that LLVM optimizes away everything if `black_box` is not used, then we should that to the users. I see two ways to do that:

- We document everything without `black_box` and add a NOTE that says that LLVM is too smart and may optimize your code away and that you may want to use `black_box` to avoid that.
- Or we do the reverse: Document everything with `black_box` and mention that it may not be necessary in some cases.

Thoughts?

japaric:
> 
    bench_function (current bench)
    bench_functions
    bench_functions_over_inputs

I like this set of names, they match the other bench methods like `bench_program` and `bench_program_with_inputs`. So perhaps this could new method could be named `bench_functions` and the future one could be named `bench_functions_with_inputs`?

src/lib.rs
faern: We make it cleaner without `black_box` so I vote for the path where we document without it and say in some caveat section that if you get ridiculous performance you should probably add some `black_box`es :)
I don't have any concrete proof or example where a `black_box` is needed on the user level, so I might be wrong, thus it's better to document without it as well. I'll fix this when I have the time.

faern:
I might be completely wrong, my mathematical mother tongue is not English, but I think `_with_input*` sounds a little bit misleading, since `bench_functions` also take an input. `_over_` sounds more like it's doing the benchmarking over a set of parameters. I'll leave that decision to you, I'll just rename the new function in this PR to `bench_functions` when I get to my computer later.

japaric:
@faern I agree that "over inputs" sounds better, but the other methods are already using "with inputs" :-/. Anyways, we don't have to take a decision right now.

faern:
Then maybe the others should be renamed to `_over_`? Let's save that discussion for another PR!

japaric:
Thanks @faern. I don't think we should rename `bench` right now, that's a breaking change as seen in the failed tests above. I'll merge this and rollback that rename. We can, later, add `bench_function` and still provide `bench` as a shorthand for it.

faern:
I don't think providing both is a good idea. That is introducing technical debt and backwards compatibility for something that is not even released yet. We have the full possibility to create something good before the first release, after that we can start thinking about compatibility.

japaric:
OK. If we are going to rename `bench` we also need to fix the tests and put a "[breaking-change]" on the commit message.

PR #65, opened 2016-02-23
Compile Doc Tests

### LINKED ISSUE #50: Mark doctests as `no_run` to avoid bitrot
They are already outdated.

killercup:
I was looking through the code and noticed there ere no tests and all doc tests were disabled. So, I fixed the doc tests and enabled them. :smile: 

Fixes #50

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -60,13 +60,14 @@ impl Bencher {
     /// # Timing loop
     ///
     /// ```rust,no_run
+    /// # use std::time::Instant;
     /// # fn routine() {}
-    /// let iters = 4_000_000;
-    /// let start = std::time::Instant::now();
+    /// # let iters = 4_000_000;
+    /// let start = Instant::now();
     /// for _ in 0..iters {
     ///     routine();
     /// }
-    /// println!("{:?}", start.elapsed());
+    /// let elapsed = start.elapsed();
     /// ```
     ///
     /// # Timing model
@@ -130,18 +131,20 @@ impl Bencher {
     /// # Timing loop
     ///
     /// ```rust,no_run
+    /// # use std::time::{Instant, Duration};
+    /// # use std::mem;
     /// # fn setup() {}
     /// # fn routine(input: ()) {}
-    /// let iters = 4_000_000;
-    /// let elapsed = std::time::Duration::new(0, 0);
+    /// # let iters = 4_000_000;
+    /// let elapsed = Duration::new(0, 0);
     /// for _ in 0..iters {
     ///     let input = setup();
     ///
-    ///     let start = std::time::Instant::now();
+    ///     let start = Instant::now();
     ///     let output = routine(input);
     ///     let elapsed_in_iter = start.elapsed();
     ///
-    ///     std::mem::drop(output);
+    ///     mem::drop(output);
     ///
     ///     elapsed = elapsed + elapsed_in_iter;
     /// }
@@ -183,17 +186,19 @@ impl Bencher {
     /// # Timing loop
     ///
     /// ```rust,no_run
-    /// let iters = 4_000_000;
+    /// # use std::mem;
+    /// # use std::time::Instant;
+    /// # let iters = 4_000_000;
     /// # fn routine() {}
     /// let outputs = Vec::with_capacity(iters);
     ///
-    /// let start = std::time::Instant::now();
+    /// let start = Instant::now();
     /// for _ in 0..iters {
     ///     outputs.push(routine());
     /// }
-    /// println!("{:?}", start.elapsed());
+    /// let elapsed = start.elapsed();
     ///
-    /// std::mem::drop(outputs);
+    /// mem::drop(outputs);
     /// ```
     ///
     /// # Timing model
@@ -229,16 +234,18 @@ impl Bencher {
     /// # Timing loop
     ///
     /// ```rust,no_run
+    /// # use std::time::Instant;
     /// # fn setup() {}
     /// # fn routine(input: ()) {}
-    /// let iters = 4_000_000;
+    /// # let iters = 4_000_000;
     /// let inputs: Vec<()> = (0..iters).map(|_| setup()).collect();
+    /// let start = Instant::now();
     ///
-    /// let start = std::time::Instant::now();
     /// for input in inputs {
     ///     routine(input);
     /// }
-    /// println!("{:?}", start.elapsed());
+    ///
+    /// let elapsed = start.elapsed();
     /// ```
     ///
     /// # Timing model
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -59,12 +59,14 @@ impl Bencher {
     ///
     /// # Timing loop
     ///
-    /// ``` ignore
-    /// start = Instant::new();
+    /// ```rust,no_run
```
japaric: What does the `rust` do here (and in all the other occurrences below)? I haven't used it before.

src/lib.rs
```diff
@@ -59,12 +59,14 @@ impl Bencher {
...
+    /// # fn routine() {}
+    /// let iters = 4_000_000;
```
japaric: Hide this variable with `#`. The number of iterations is chosen by the `Bencher` at runtime, this gives the impression that `iters` is fixed.

src/lib.rs
```diff
@@ -59,12 +59,14 @@ impl Bencher {
...
+    /// let start = std::time::Instant::now();
```
japaric: I think it would be better to import `std::time::Instant` to avoid the full path here.

src/lib.rs
```diff
@@ -59,12 +59,14 @@ impl Bencher {
...
     /// for _ in 0..iters {
     ///     routine();
     /// }
-    /// elapsed = start.elapsed();
+    /// println!("{:?}", start.elapsed());
```
japaric: revert this back to the `elapsed` binding because the "Timing model" section below refers to it.

src/lib.rs
```diff
@@ -127,18 +129,21 @@ impl Bencher {
     ///
     /// # Timing loop
     ///
-    /// ``` ignore
-    /// elapsed = 0;
+    /// ```rust,no_run
+    /// # fn setup() {}
+    /// # fn routine(input: ()) {}
+    /// let iters = 4_000_000;
```
japaric: hide this for the same reason stated above.

src/lib.rs
```diff
@@ -127,18 +129,21 @@ impl Bencher {
...
+    /// let elapsed = std::time::Duration::new(0, 0);
```
japaric: as above, importing `Duration` would remove the need to use a full path here.

src/lib.rs
```diff
@@ -127,18 +129,21 @@ impl Bencher {
...
     /// for _ in 0..iters {
     ///     let input = setup();
     ///
-    ///     let start = Instant::now();
+    ///     let start = std::time::Instant::now();
```
japaric: ditto, import - full path.

src/lib.rs
```diff
@@ -127,18 +129,21 @@ impl Bencher {
...
     ///     let output = routine(input);
     ///     let elapsed_in_iter = start.elapsed();
     ///
-    ///     mem::drop(output);
+    ///     std::mem::drop(output);
```
japaric: ditto, import `std::mem` - use `mem::drop`.

src/lib.rs
japaric: hmm, shouldn't `elapsed` need to be mutable because of the re-assignment below?

src/lib.rs
```diff
@@ -177,16 +182,18 @@ impl Bencher {
     ///
     /// # Timing loop
     ///
-    /// ``` ignore
-    /// outputs = Vec::with_capacity(iters);
+    /// ```rust,no_run
+    /// let iters = 4_000_000;
```
japaric: ditto. hide.

src/lib.rs
```diff
@@ -177,16 +182,18 @@ impl Bencher {
...
+    /// # fn routine() {}
+    /// let outputs = Vec::with_capacity(iters);
     ///
-    /// start = Instant::now();
+    /// let start = std::time::Instant::now();
```
japaric: ditto. import, full path.

src/lib.rs
```diff
@@ -177,16 +182,18 @@ impl Bencher {
...
     /// for _ in 0..iters {
     ///     outputs.push(routine());
     /// }
-    /// elapsed = start.elapsed();
+    /// println!("{:?}", start.elapsed());
```
japaric: ditto, revert to `elapsed` assignment.

src/lib.rs
```diff
@@ -177,16 +182,18 @@ impl Bencher {
...
     ///
-    /// mem::drop(outputs);
+    /// std::mem::drop(outputs);
```
japaric: ditto. import `std::mem`.

src/lib.rs
```diff
@@ -221,14 +228,17 @@ impl Bencher {
     ///
     /// # Timing loop
     ///
-    /// ``` ignore
-    /// inputs = (0..iters).map(|_| setup()).collect();
+    /// ```rust,no_run
+    /// # fn setup() {}
+    /// # fn routine(input: ()) {}
+    /// let iters = 4_000_000;
```
japaric: ditto. hide.

src/lib.rs
```diff
@@ -221,14 +228,17 @@ impl Bencher {
...
+    /// let inputs: Vec<()> = (0..iters).map(|_| setup()).collect();
     ///
-    /// start = Instant::now();
+    /// let start = std::time::Instant::now();
```
japaric: ditto. import.

src/lib.rs
```diff
@@ -221,14 +228,17 @@ impl Bencher {
...
     /// for input in inputs {
     ///     routine(input);
     /// }
-    /// elapsed = start.elapsed();
+    /// println!("{:?}", start.elapsed());
```
japaric: ditto. `elapsed` assignment.

japaric:
Thanks for the PR! The "timing loops" were more like pseudo-code in my mind, but I think it's a great idea to test them :+1:. I left a bunch of comments :smiling_imp:.

src/lib.rs
killercup: It's just Markdown's syntax to set the language used in the code block (just like `text` below). I like to add it to make things more explicit. The `ignore` or `no_run` flags only work when `rust` is set of course – rustdoc assumes `rust` by default when no language is set.

src/lib.rs
killercup: I'm gonna change this to (hidden) imports – that's what you meant, right?

killercup:
I think I addressed all your comments :smiley:

I originally wanted to use full paths everywhere so people knew this was from `std`, by the way. I don't think that's really necessary, though, so I changed it to (hidden) imports.

japaric:
Thanks! :+1:

PR #69, opened 2016-02-24
accept `Duration` instead of `u64` nanos in `Criterion`'s setter methods

japaric:
[breaking-change]

Change your code from:

``` rust
Criterion::default().warm_up_time(1_000_000_000)...
```

to:

``` rust
Criterion::default().warm_up_time(Duration::from_secs(1))...
```


japaric:
cc @faern

PR #75, opened 2016-02-29
settle on API for release

japaric:
[breaking-change] the following functions have been renamed:

- `Criterion::bench` -> `Criterion::bench_function`
- `Criterion::bench_with_inputs` -> `Criterion::bench_function_over_inputs`
- `Criterion::bench_program_with_inputs` -> `Criterion::bench_program_over_inputs`

---

r? @faern


faern:
I like the name changes. However, we still have `analysis::program_with_inputs` and similar. Should we not also rename the internal functions while at it? That does not change the API but anyway.

PR #77, opened 2016-03-13
Add clippy

bsteinb:
- Add `clippy` to the list of dependencies.
- Add a `lint` feature that enables `clippy` for now.
- Appease `clippy`'s lints.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,9 +36,11 @@ before_script:
 
 install:
   - sh ci/install.sh
+  - export PATH=$HOME/cargo-clippy/target/release:$PATH
 
 script:
   - cargo build
+  - cargo clippy
   - cargo test
   - cargo bench -- --test --nocapture
   - cargo doc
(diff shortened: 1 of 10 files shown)


japaric:
Thanks for the PR!

I'm pretty much :+1: on fixing the clippy warnings but I would prefer to run [cargo-clippy] within Travis CI instead of adding clippy as a Cargo dependency. Let me know if you want to tackle the Travis CI part, otherwise I'd be happy to only merge the second commit in this PR.

[cargo-clippy]: https://github.com/arcnmx/cargo-clippy

bsteinb:
I have gotten cargo-clippy to work. If this is to your taste I will try to do the same for stats.rs.

japaric:
> I have gotten cargo-clippy to work.

Excellent! Thank you. I'll merge this after appveyor finishes its tests.

> If this is to your taste I will try to do the same for stats.rs.

I would appreciate that :smile:.

PR #78, opened 2016-04-12
remove stable `time2` feature

llogiq:
Since that feature is stabilized, we no longer need the feature gate.


japaric:
Thanks!

llogiq:
You're welcome.

PR #80, opened 2017-09-10
Fix compile errors in macros, allowing tests to be run successfully.

bheisler:
With this change and https://github.com/japaric/simplot.rs/pull/47, all three criterion crates build and complete all tests on the latest Nightly build (at least on my computer. May need some more work to make Travis/Appveryor happy).


japaric:
Thanks, @bheisler.

PR #98, opened 2017-12-20
switch to serde from rustc_serialize

maccam912:
As mentioned in https://github.com/japaric/criterion.rs/issues/91


bheisler:
Looks good to me. Thanks!

PR #100, opened 2017-12-22
Replace floaty dependency with num-traits 

### LINKED ISSUE #95: Replace Floaty with num-traits
Floaty doesn't seem to be maintained anymore. num-traits seems to be the standard for this sort of thing. I think it should just be a matter of updating all of the trait bounds and imports in stats.

Difficulty: Easy

babbageclunk:
At the moment num_traits::float::Float doesn't have Sync or Send, and conversion from a primitive always returns an Option. Add a small Float trait that adds more convenient safe casting and can be used with threads.

Also fixes some small clippy warnings.

Fixes #95.

### DIFF
--- a/src/analysis/mod.rs
+++ b/src/analysis/mod.rs
@@ -43,7 +43,7 @@ pub fn summarize(id: &str, criterion: &Criterion) {
 pub fn function<F>(id: &str, f: F, criterion: &Criterion) where F: FnMut(&mut Bencher) {
     common(id, &mut Function(f), criterion);
 
-    println!("");
+    println!();
 }
 
 pub fn functions<I>(id: &str,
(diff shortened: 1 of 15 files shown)


babbageclunk:
I don't understand what's causing that failure - it looks like nightly has removed `lifetimes` from `syntax::ast::BareFnTy`?

bheisler:
Wonderful! This is excellent work, much appreciated.

The build failure looks unrelated - incompatibility between quickcheck-macros and the latest nightly compiler.

bheisler:
Thanks for the contribution! If you'd like to stick around, feel free to take a look at some of the other issues, or try benchmarking something and let me know what you think.

PR #102, opened 2017-12-23
Don't panic on errors

### LINKED ISSUE #99: Handle File IO Errors
We should handle file IO errors and encode/decode failures instead of just panicking. Fortunately, we don't need to bubble errors all the way up to the public API, since the user wouldn't be able to do anything with them anyway.

For now, let's just focus on the panics and unwraps in the `src/fs.rs` module. There are a bunch of panics in the plots code as well, but those can wait.

Failure to load data from a file should be treated as if that file didn't exist. That is, in `src/analysis/compare.rs` we should print an appropriate error message and proceed without doing the comparison. In `src/plot/mod.rs`, we should print a message and skip that particular sample. In either case, this will require bubbling the Result from `fs::load` up to the caller and maybe one level above that.

If we fail to write data to disk we can just print a message and carry on.

torkleyy:
What's left is to initialize a logger implementation.

---

Fixes #99

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Fixed
+- Don't panic on IO errors
 
 ## [0.1.1]
 ### Added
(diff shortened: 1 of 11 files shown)


bheisler:
Thanks for the pull request. I probably won't get to it right away with the holidays and everything, but from a quick glance it looks good.

For now, could you rebase on the latest changes? I've removed the dependency on quickcheck-macros which is causing the builds to fail.

bheisler:
Merged. Thanks for the help!

PR #114, opened 2018-01-13
Minor book additions for migration from libtest

paupino:
Added step for adding `dev-dependencies` and also appended missing `!` for macro invocations.

On a side note, I see that the line endings may have been changed as a result. I'm happy to change these to whatever is preferred however may I suggest also adding an [`.editorconfig`](http://editorconfig.org/)? This will help maintain consistency going forward.


bheisler:
Thanks!

I didn't know about editorconfig, I'll probably add one of those soon.

PR #117, opened 2018-01-18
Update dependencies

alexbool:


bheisler:

Could you maybe give me an idea of what the goal is here? I'm not entirely against updating dependencies, but I'd prefer not to until there's a specific reason to do so. Since Criterion.rs isn't security-critical there isn't as much need to keep things up to date. Is there some problem with the existing dependency versions that I'm not aware of?

Also, please don't increment the Criterion.rs version numbers. I'll do that myself when it's time to make a new release.

alexbool:
Hi @bheisler!
I use `criterion-plot` in one of my projects. (And BTW it is the best plotting library for rust I have found.)
The motivation to keep up-to-date dependencies comes from fact that many versions of same dependencies (including private ones) can seriously increase compilation times and even resulting binary size. So often it is better to have freshest verisions of common libraries (like itertools) possible.
Sorry, I should have menthioned this in the PR description.

Regarding bumping version numbers, I will revert them shortly.

bheisler:
It looks like there are still some uses of `itertools::linspace` that will need to be updated. Otherwise this looks good.

Although of course you're free to use `criterion-plot` if you wish, you should be aware that I consider it to be an internal implementation detail of Criterion.rs - I might make breaking changes without warning.

alexbool:
Updated.

bheisler:
Merged. Thanks for the pull request!

benesch:
For posterity, if anyone else is thinking about submitting a similar dependency bump PR, note that Criterion.rs now has a policy of bumping all dependencies prior to a release, and does not accept dependency bump PRs otherwise: https://github.com/bheisler/criterion.rs/blob/master/CONTRIBUTING.md#a-note-on-dependency-updates.

PR #143, opened 2018-05-02
deps: Update handlebars.

waywardmonkeys:
Moving to handlebars 0.32.1 brings an update to the regex dependency
to 1.0.


bheisler:
Thanks for the pull request!

Issue #147, opened 2018-05-23
Csv Report


nickbabcock:
While the criterion html report is decent for most situations, there are occasions when a different graph allows for an easier interpretation. Instead making the html report endlessly configurable, I propose exposing the data in a csv format. This way, anyone can easily create their graph in R, python, etc. While the json in `target/criterion/$BENCHMARK/$FUNCTION` is a good start -- pulling everything together into a (preferably) single file would greatly lower the bar.

For instance a `$cargobenchname-raw.csv` could hold the raw sampling data with headers like:

| group id | id | parameter | elapsed (ns) |
| -------- | -- | --------- | ------------ |
| a           | b  |                | 100             |
| a           | b  |                | 150             |

Where group id and parameter can be left blank for the benchmarks that don't apply. I'm torn if throughput should be included. I'm leaning towards no, as I believe it's calculable. Users of the raw data can calculate additional aggregations they are interested in (like percentiles)

Alternatively (or in addition), the aggregated values can be written to `$cargobenchname.csv` 

| group id | id | parameter | elapsed (ns) | x̅ | x̅ p95 upper | x̅ p95 lower | σ<sub>x̅<sub> |
| -------- | -- | --------- | ------------ | - | ----------- | ----------- | -- |

More than just the mean statistics would be included.

As a comparison, the [C# BenchmarkDotNet](https://github.com/dotnet/BenchmarkDotNet) outputs a [csv like this](https://github.com/nickbabcock/Farmhash.Sharp/blob/20888cde1692f89d2bae059034471ea85e15593a/benchmarks.csv), which can be slurped up to [create nice looking graphs](https://nickbabcock.github.io/Pfim/img/benchmark.png).

Maybe this could help people transition to presenting rust benchmark results in graphical form. I know that my Rust projects don't get enough love in the dataviz department 😄

bheisler:
Hey, thanks for the suggestion and sorry for the slow response.

There have been requests for a stable machine-readable output format (eg. for [lolbench](https://github.com/anp/lolbench)). The current JSON storage format is an internal implementation detail. I was leaning more towards defining a stable JSON interchange format that could be different from the storage format, rather than CSV, but I could be convinced otherwise. Why did you choose to go with CSV over JSON?

[...]

PR #148, opened 2018-05-24
Add csv report

nickbabcock:
This PR adds a CSV report that contains all the sampling data across all benchmarks in a `benchmark-raw.csv`. It appears like the following:

| group | function | parameter | elapsed |
| -------- | -- | --------- | ------------ |
| a           | b  |                | 100             |
| a           | b  |                | 150             |

Like `html_reports`, `csv_reports` is feature that is enabled by default.

Some thoughts:

- Assuming that adding `Report::report_init` is necessary (to clear out file and add csv header) -- what is your name preference? `report_start` is another possibility ... or some other name
- I created a default implementation for `report_init` for all reports. I noticed that this goes against the current style, but I couldn't think of why all the `Report` impls need to define all the functions (instead of relying on the default empty function).
- Ideally if I'm running `bencha.rs` then the csv would be `bencha-raw.csv`, but I didn't see that info available (I could be missing something). This may not be a big issue as I only tend to see projects use only a single file for benchmarks.

I've tested this on some of my crates and things looks great

EDIT: related #147

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,7 @@ serde_json = "1.0"
 serde_derive = "1.0"
 atty = "0.2"
 clap = "2.29"
+csv = { version = "1", optional = true }
 handlebars = { version = "0.32", optional = true }
 
 [dev-dependencies]
@@ -40,7 +41,8 @@ maintenance = { status = "passively-maintained" }
 [features]
 real_blackbox = []
 html_reports = ["handlebars", "criterion-plot"]
-default = ["html_reports"]
+csv_reports = ["csv"]
+default = ["html_reports", "csv_reports"]
 
 [workspace]
 
--- a/src/benchmark.rs
+++ b/src/benchmark.rs
@@ -597,6 +597,7 @@ where
 
         let mut any_matched = false;
 
+        c.report.report_init(&report_context);
         for routine in self.routines {
             for value in &self.values {
                 let function_id = if num_routines == 1 && group_id == routine.id {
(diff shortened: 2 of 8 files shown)


bheisler:
Hey, thanks for the pull request, and sorry for the slow response.

Unfortunately, this still needs some work.

* First, even running the unit tests with these changes causes a bunch of IO errors to be printed. This is probably because you haven't added a `report_init` call to `ParameterizedBenchmark::run`.
* Unfortunately, this approach in general won't work. If there is more than one benchmark defined in a suite, all subsequent benchmarks will overwrite the file generated by the previous ones. You'll need to write a separate CSV file for each benchmarked function like the HTML reports do. The reason it will need to be per-function rather than per-group is because the user can filter which benchmarks are run on the command-line. Happily, this removes the need for the `report_init` call entirely. You can use Criterion.rs's own example benchmarks to test this.
* As for the default implementation, the reason that I didn't do that for the other functions is that I completely forgot that you could do that. My bad.

Also, please see comments on the related issue.

bheisler:
Hey, thanks for the pull request and for the discussion.

Ultimately I've decided to implement this myself, as I wanted to take this feature in a slightly different direction and go with the file-per-benchmark approach (at least for the first iteration). If this presents a problem for your use case, we can consider how to handle that later on.

nickbabcock:
No problem; happy to pitch my ideas 😄 

I'll test out this implementation when I get a chance and report back.

PR #153, opened 2018-06-15
Add --no-overwrite option

### LINKED ISSUE #150: Avoid overwriting the previous run
When I try to optimize something using the normal rust benchmarks I usually do an initial baseline run of the code before I start and then save the output in a file somewhere. Then, after making the (supposed) improvements I rerun the benchmarks again and compare the results with the baseline to see if it is has improved.

While criterion has this comparison built (which is great!) it also always overwrites the old data after each run which makes it impossible (well, really annoying) to compare against a fixed baseline. I saw the `--measure-only` flag which disables this overwriting but but unfortunately that does not display any analysis at all.

Perhaps a flag which just disables the overwriting on disk could be added?

Everlag:
This disables saving results to disk post-analysis.

This is useful when iterating against a baseline(eg. comparing a feature branch to the commit it branches off master)

Typical use is an initial `cargo bench` to build a baseline then every future bench using `--no-overwrite`.

Specifically, this introduces two new attributes on Criterion and ReportContext, `no_overwrite`. Whenever Criterion is associated with a ReportContext, `no_overwrite` is copied over. This flag indicates to all reports that they should never modify the filesystem.

An associated test, `test_without_overwrite`, has been added. This ensures that the previous results are neither mutated nor removed.

Closes #150

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+- Added a `--no-overwrite` argument which causes the benchmark executable to run
+  the  warmup, measurement, and analysis then move on to the next benchmark without
+  saving data. This is useful when iterating against a baseline
+  (eg. comparing a feature branch to the commit it branches off master)
 
 ## [0.2.3]
 ### Fixed
--- a/src/analysis/mod.rs
+++ b/src/analysis/mod.rs
@@ -54,8 +54,6 @@ pub(crate) fn common<T>(
 
     criterion.report.analysis(id, report_context);
 
-    rename_new_dir_to_base(id.id(), &criterion.output_directory);
-
     let avg_times = iters
         .iter()
         .zip(times.iter())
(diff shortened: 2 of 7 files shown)


bheisler:

I'm not sure that this is the right approach. For one thing, I think users would want to see the HTML reports updated with the latest comparison data. For another, I think this will cause Criterion.rs to compare against the wrong data.

In the normal process, Criterion.rs keeps data in "new" and "old" directories. If there is data in the "new" directory, it moves it to "old" (overwriting what was there) and then writes fresh data into the "new" directory. Then it loads data from the "old" directory to use for the comparison. In short, after a Criterion.rs run, the "new" directory contains data from the last run, while the "old" directory contains data from two runs ago.

With this change, we skip the move-and-overwrite step, which will cause Criterion.rs to load the data from the "old" directory (ie. two runs ago) to use as the baseline for the comparison. I don't think that's what the user would expect. What you probably want is to always do the move-and-overwrite and potentially skip writing the "new" data, but then you have the added complexity of making sure Criterion.rs can handle having "old" data but not "new".

However, you've got me thinking about alternate ways to meet this goal, and I think I've come up with a good one. Suppose we add two new arguments instead of one (`--save-baseline <name>` and `--baseline <name>` perhaps). These would be mutually-exclusive.

* If the user passes `--save-baseline`, Criterion.rs performs the measurements and comparison as usual, writes the "new" values (including all reports), and at the end it copies-and-overwrites (not moves) the "new" data & reports to a directory named after the baseline name. It could even use that directory as the baseline for the comparison in this run, which reduces the special-cases and is probably what the user wants anyway (when I'm updating an existing baseline, I probably want to see the comparison between the new and old values).
* If the user passes `--baseline` then Criterion.rs loads the baseline values for the comparison from the named baseline directory (failing early and loudly if it doesn't exist). In this situation, it will not perform the copy-and-overwrite at all, but will still overwrite "new".
* If the user passes neither, then Criterion.rs behaves as if `--save-baseline old` was given.

This has a few advantages. For the user, it allows them to be extremely clear about what they want, and it allows them to manage multiple different baselines (eg. for multiple parallel development branches or comparing multiple different time scales). For the developers, it neatly reduces the problem to some simple file-path manipulation. It does have a slight quirk that after a run the "old" data is equal to the "new" data but that's probably not a big deal.

What do you think of this idea?

Everlag:
I think the `--baseline`/`--save-baseline` feature set is less fragile and more usable than `--no-overwrite`. 

I'm closing this PR in favor of working on baselines. Having worked on this change, I think the path modification approach will be much simpler than forcing conditionals into every report. I'll ping you on this thread if I have any questions about intended semantics.

Thanks for the review.

PR #161, opened 2018-06-27
Link to image for violin plot

djc:
Improves consistency with other plots on top of being generally useful.


bheisler:
Thanks again!

PR #165, opened 2018-06-30
Improve error message when loading estimates file

KillTheMule:
Closes https://github.com/japaric/criterion.rs/issues/164.

I went for the least intrusive approach, since I don't know the codebase very well. Since you're using failure, it was very easy to add a bit of context here.

I know `context` with `format` has a bit of a performance penalty, but for `with_context` one needs a `Fail` rather than a `failure::Error`, and I'm not sure how to arrange that or if that's even what one wants. Otoh, when loading a file, an additional `format` is probably dwarfed by everything else...


bheisler:

PR #175, opened 2018-07-16
fix links

hrls:


bheisler:
Thanks for the pull request! Looks like I forgot to update that.

PR #180, opened 2018-07-18
sort sub-benchmarks by name (fixes #179)

### LINKED ISSUE #179: in HTML report, sort results of parameterized benchmarks in the order they're run
In command line output, parameterized benchmark results are displayed in the order in which the benchmarks are run, but in HTML reports they aren't.

For example, the parameterized benchmarks in [mykmelez/kvbench](https://github.com/mykmelez/kvbench/) are run in the order:

* (1, 1)
* (1, 100)
* (1, 1000)
* (100, 1)
* (100, 100)
* (100, 1000)
* (1000, 1)
* (1000, 100)
* (1000, 1000)

And that's how they're displayed in the command-line output:

```
leveldb_put_seq_sync/(1, 1)
                        time:   [54.607 us 55.391 us 56.427 us]
leveldb_put_seq_sync/(1, 100)
                        time:   [56.160 us 57.147 us 58.843 us]
leveldb_put_seq_sync/(1, 1000)
                        time:   [72.128 us 73.193 us 74.640 us]
leveldb_put_seq_sync/(100, 1)
                        time:   [202.97 us 217.77 us 235.79 us]
leveldb_put_seq_sync/(100, 100)
                        time:   [548.61 us 559.57 us 568.20 us]
leveldb_put_seq_sync/(100, 1000)
                        time:   [626.91 us 632.08 us 637.58 us]
leveldb_put_seq_sync/(1000, 1)
                        time:   [1.2474 ms 1.2581 ms 1.2680 ms]
leveldb_put_seq_sync/(1000, 100)
                        time:   [1.3897 ms 1.3970 ms 1.4058 ms]
leveldb_put_seq_sync/(1000, 1000)
                        time:   [6.2530 ms 6.3345 ms 6.4014 ms]
```

(Which makes sense, since Criterion.rs is presumably writing the results to stdout after each run.)

But in [this HTML report](https://mykmelez.github.io/kvbench/criterion/report/) from that project, they're displayed in what seems at first glance to be a random order:

* (1000, 1000)
* (100, 1)
* (1, 1000)
* (1, 1)
* (100, 100)
* (100, 1000)
* (1000, 100)
* (1000, 1)
* (1, 100)

That makes it harder to understand the parameterization model and to compare results across step changes in the parameters (assuming the benchmark authors ordered them appropriately, which is within their control). It'd be more usable to sort those results in the order they're run, just as with the command line output.

mykmelez:
This sorts the sub-benchmarks by name when generating an HTML report, resolving #179 (assuming you agree that we should fix that! I know that this PR is putting the cart before the horse, as we haven't had any discussion in that issue yet, but I figured it would be a small change, and I wanted to familiarize myself with the codebase).

Note that [CONTRIBUTING](https://github.com/japaric/criterion.rs/blob/master/CONTRIBUTING.md) suggests to run `cargo +nightly fmt`, but that currently fails for me with:

```
thread 'main' panicked at 'not yet implemented', tools/rustfmt/src/expr.rs:346:47
```

Which is https://github.com/rust-lang-nursery/rustfmt/issues/2743. However, I don't foresee any formatting issues with this small (two-line) change. And in any case rustfmt makes a bunch of other changes to the code before panicking, so it looks like the codebase hasn't been getting formatted lately (or at least not with recent versions of rustfmt!).

### DIFF
--- a/src/html/mod.rs
+++ b/src/html/mod.rs
@@ -308,10 +308,11 @@ impl Report for Html {
                 .file_name()
                 .map(|name| name.to_string_lossy())
                 .unwrap();
-            let sub_benchmarks = fs::list_existing_reports(&base_dir)?
+            let mut sub_benchmarks = fs::list_existing_reports(&base_dir)?
                 .into_iter()
                 .map(|sub_path| path_to_individual_benchmark(&sub_path, output_directory))
                 .collect::<Result<Vec<_>>>()?;
+            sub_benchmarks.sort_unstable_by_key(|k| k.name.clone());
             Ok(IndexBenchmark {
                 name: name.to_string(),
                 path: path.join("index.html")


bheisler:
Thanks for the pull request!

PR #181, opened 2018-07-19
deps: Update to handlebars 1.0.

waywardmonkeys:


waywardmonkeys:
The failed build seems to be due to `handlebars` 1.0 not supporting Rust 1.23 ... I suspect this should be improved upstream as it doesn't seem like a necessary breakage.

bheisler:

Could you explain the reasoning behind this change? Technically, bumping the minimum required Rust version is allowed under the compatibility policy, but I don't want to inconvenience existing users without a good reason. Handlebars 0.3 has been working fine as far as I know.

waywardmonkeys:
It was fine ... I just noticed in another project that it built `handlebars` twice because it was using the 1.0 already. But I'm not bothered if we just close this.

bheisler:
In that case, I think I'll just close this for now. Thanks anyway though!

PR #196, opened 2018-08-27
Get Criterion building with `--minimal-versions`

### LINKED ISSUE #183: Add CI build with `--minimal-versions`
Should verify that Criterion.rs builds correctly with this flag, to ensure that the lower-bounds for dependencies are correct.

Eh2406:
This is a start on #183. It gets Criterion to build with `--minimal-versions`! It has been a long road to get here, but in the process a large number of your dependencies got working with it.

In order to close #183 we would need to add it to CI. Do to the complexity of your CI I did not see how to do that. Suggestions are welcome. Pryor art includes: [Cargos CI](https://github.com/rust-lang/cargo/blob/master/.travis.yml#L29-L40) and [walkdir CI](https://github.com/rust-lang/cargo/blob/master/.travis.yml#L29-L40)


bheisler:
Awesome, thanks! I can handle the CI updates.

I notice that you've updated `handlebars` - is this necessary for the minimal-versions build to work? The reason I ask is because that version of `handlebars` (or more specifically, the transitive dependency on `pest`) doesn't build on Rust 1.20. Technically, Criterion.rs doesn't have to maintain compatibility with 1.20, but I don't want to break that compatibility if I don't have to.

Eh2406:
That is a frustrating situation. The update to handlebars was made to include this [PR](https://github.com/sunng87/handlebars-rust/pull/231) and each line of that diff is its own PR including [this PR to pest_derive](https://github.com/pest-parser/pest/pull/255).

bheisler:
Ah, I see. That's unfortunate.

In that case, I think I'd rather put this on hold for now. I don't think it makes sense to break compatibility over something like minimal-version support, but I will definitely merge this in next time I have to break compatibility for some other reason.

Thanks again for the help!

Eh2406:
The problems with 1.23 [should](https://github.com/sunng87/handlebars-rust/pull/231#issuecomment-416790944) be fixed now. closing and reopening to trigger new CI.

bheisler:
Looks good to me! Thanks for all the hard work.

PR #197, opened 2018-08-30
Unindent `Cargo.toml` additions

sunjay:
Typically sections in Cargo.toml do not start out indented. By using both code block syntax and indenting 4 spaces, you have inadvertently made it so that people copying and pasting those lines end up with something indented that shouldn't be indented.

This change fixes that.


bheisler:
Fair enough! Thanks for the pull request.

PR #199, opened 2018-09-02
Update quickcheck to 0.7

ignatenkobrain:


bheisler:

Could you explain why you feel this change is needed?

ignatenkobrain:
In Fedora we try to package only latest version so this would help us with packaging efforts. Debian does the same.

Is there any reason to hold it on old version?

bheisler:
No, I just generally prefer to be lazy and update dependencies as-needed rather than doing it whenever there's a new version available.

Marged. Thanks!

PR #202, opened 2018-09-12
Update approx to 0.3

ignatenkobrain:


bheisler:
I've updated the dependencies, including updating approx to 0.3. Thanks for the pull request anyway!

PR #204, opened 2018-09-12
Update num-complex to 0.2

ignatenkobrain:


bheisler:
Thanks for the pull request. 

Looks like this is breaking the CI build. You'll need to fix up the code before I can merge it.

cuviper:
You have disabled default-features, and num-complex 0.2 added a "std" feature gating some methods.  You can either enable the "std" feature explicitly, or just enable default features.

bheisler:
I've updated the dependencies, including updating num-complex. Thanks for the pull request anyway, though!

Issue #195, opened 2018-08-27
Remove Unsafe Blocks


bheisler:
There are a lot of old `unsafe` blocks in Criterion.rs' internals. Many of them were written to work around compiler bugs or missing features that have since been fixed (see https://github.com/japaric/criterion.rs/issues/188#issuecomment-416001204). These should be cleaned up.

There are a few others that might be harder to remove without impacting analysis performance in the bootstrapping code. I've tried several times to remove these without much success so far.

gnzlbg:
I managed to remove many unsafe blocks in #208 . Most of the remaining ones are due to the usage of `thread_scoped`, here we should migrate criterion either to `crossbeam::scope`, or maybe even just `rayon`. 

The remaining `unsafe` blocks are only in the `stats` crate and hint about the crate's problematic design. For example, some transmutes are just used to workaround some types not having `new` methods (to subvert privacy boundaries), others are also used to subvert the type system in weird ways: for example, `Sample<A: Float>` but `Distribution<A>` does not require `A: Float`, yet distributions with `A`s that are not floats are transmuted into `Sample`s.

cuviper:
> or maybe even just `rayon`. cc @cuviper thoughts?

How could I say no? 😉   But yes, indeed, those look like prime candidates for parallel iterators.

If you wanted to get fancy, those `Builder`s could probably implement `FromParallelIterator` to directly fill their target vectors, but it's probably good enough to leave it with `sub_distributions` and the `extend` merging at first.

bheisler:
I've tried using Rayon there, in fact. It was a while ago, though, so it might work better now. The problem I ran into is that, well, some of that code is super complex (I'm looking at `mixed.rs` here). There is a thread-local structure (`Resamples`) which holds a borrow of the shared vector `c`. I couldn't figure out a way to do that in Rayon without creating a new `Resamples` instance for every iteration (potentially expensive because it contains a random number generator as well, which needs to be initialized from the `thread_rng()`) or crossing the borrow checker or both. I believe I also tried to implement `FromParallelIterator` for the Builders and gave up eventually, though I don't remember why. The other obvious approach - using `split_at_mut` and giving each chunk to a different scoped thread - also doesn't work, because of the way it tries to be generic over the number of output vectors.

[...]

PR #208, opened 2018-09-20
Flyby cleanups

### LINKED ISSUE #203: include LICENSE-* in sub-crates
Unfortunately tarballs on crates.io do not contain it while Apache-2.0 requires that.

Easiest way would be just to create symlinks from appropriate subdirectories.

Thanks for cooperation!

gnzlbg:
* tidy the `ci` scripts a bit
* bump minimum supported rust version to 1.25.0 (probably a breaking change)
* add licenses to the stats and plot libs (closes #203)
* update rustfmt and clippy
* update dependencies
* implement Deref for Sample
* remove some usages of unsafe code (helps with #195)


bheisler:
Wow, this is great! Thanks a bunch for doing all of this.

Overall, it looks good. I'm concerned about increasing the minimum Rust version, though. Technically, the compatibility policy says we can increase the minimum supported version to 1.26 (though I would prefer to maintain compatibility with 1.23 if it's reasonably practical). You've increased the minimum version to 1.28, though. I'm afraid that won't work. Can you explain why you've increased the minimum version?

The CI failure can be ignored - that test is a bit flaky.

gnzlbg:
> Can you explain why you've increased the minimum version?

So I bumped it to 1.25 first, but once I start removing unsafe code I discovered that `Sample` was invoking undefined behavior because it is not `repr(transparent). Rust 1.28 is the minimum Rust version that supports `repr(transparent)` on stable.

It might be possible to avoid undefined behavior in `Sample` by re-designing the api, e.g., if we change it to `Sample<'a, T>(&'a [T])` instead of the current `&Sample<T>([T])`, but that's going to be a greater breaking change.

bheisler:
Yeah, I'm right there with you. I don't really understand why Sample was designed as it was either - it was like that when I took over maintenance. I suppose I should have refactored a bunch of that stuff.

Regardless, we can't use the `repr(transparent)` tag for a while yet without breaking compatibility. I understand that it's UB, but it's UB that happens to work for now and will probably continue to do so for the near future. I'd prefer to go ahead and convert Sample into a genuinely safe API - `criterion-stats` isn't public API so I don't care about breakage there - but perhaps that should be a separate pull request. For the time being, I'd say just leave it as-is and maybe add a comment to the unsafe spots.

gnzlbg:
Sounds good, I'll ping you when its done. FWIW I think that if one is going to refactor this, the best way would be to just drop criterion stats completely and use one of Rust statistics libraries. They might need some work to support some criterion stats use cases, but that's probably time well spent since other people will take on the maintenance of those parts for you.

bheisler:
> FWIW I think that if one is going to refactor this, the best way would be to just drop criterion stats completely and use one of Rust statistics libraries.

Yeah, that would be nice. I haven't looked into other statistics libraries yet. `criterion-plot` should probably be spun out into its own library as well, if I could find someone who wanted to maintain it.

bheisler:
Thanks for the help!

Rather than make you fix all of the merge conflicts, I've cherry-picked some of your changes onto master.

I decided not to bump the minimum version, since I'm planning to do a breaking-change release soon and I figure I might as well get all of the breakage done at once.

PR #209, opened 2018-09-20
Support custom test frameworks behind ctf feature

### LINKED ISSUE #205: use custom_test_frameworks on nightly
Instead of doing the whole `bench!`, `criterion_main!`, etc. on nightly, it would be nice if `criterion` would just use the `custom_test_frameworks` feature: https://github.com/rust-lang/rust/blob/master/src/doc/unstable-book/src/language-features/custom-test-frameworks.md

gnzlbg:
When the `--feature=ctf` is enabled, this PR exports a custom test runner and a `#[criterion]` proc macro from the `criterion` crate. This allows using criterion without the `criterion_group!` and `criterion_main!` macros by just using the `#[criterion]` proc macro instead (see `tests/ctf.rs` for a full example):

```rust
#[criterion]
fn compare_fibonaccis(c: &mut Criterion) {
    let fib_slow = Fun::new("Recursive", |b, i| b.iter(|| fibonacci_slow(*i)));
    let fib_fast = Fun::new("Iterative", |b, i| b.iter(|| fibonacci_fast(*i)));

    let functions = vec![fib_slow, fib_fast];

    c.bench_functions("Fibonacci", functions, 20);
}
```

Currently, each function marked with the `#[criterion]` attribute becomes its own benchmark group, but we might want to improve this in the future, e.g., using modules to express benchmarks groups, and functions to express benchmarks, like:

```rust
#[criterion_group]
mod fibonacci {
    #[criterion_benchmark]
    fn slow(c: &mut Criterion) { ... }

    #[criterion_benchmark]
    fn fast(c: &mut Criterion) { ... }
}
```

Closes #205 .

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,7 @@ backtrace = "0.3.5" # required only for minimal-versions. brought in by failure.
 chrono = "0.4.6" # required only for minimal-versions. brought in by simplelog.
 criterion-plot = { path="plot", version="0.2.5", optional = true }
 criterion-stats = { path="stats", version="0.2.5" }
+criterion-macro = { path="macro", version="0.2.5", optional = true }
 failure = "0.1.2"
 failure_derive = "0.1.1"
 itertools = "0.7"
@@ -44,6 +45,7 @@ maintenance = { status = "passively-maintained" }
 real_blackbox = []
 html_reports = ["handlebars", "criterion-plot"]
 default = ["html_reports"]
+ctf = [ "criterion-macro" ]
 
 [workspace]
 
--- a/macro/Cargo.toml
+++ b/macro/Cargo.toml
@@ -1,0 +1,13 @@
+[package]
+name = "criterion-macro"
+version = "0.2.5"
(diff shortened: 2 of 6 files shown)


bheisler:
Hey, thanks for the pull request. I wish you'd spent a bit more time talking to me before doing all this work, though. I wasn't planning to do this right away - I intend to use the switch to test-frameworks as an excuse to do a pretty major redesign of Criterion.rs and its public API. I don't have the bandwidth available to do that properly right now, and I don't think I can guide someone else to do it because I haven't really figured out exactly what I want to do yet. I'm sorry to say this after you've done all this hard work, but If I merge this in, there's a pretty good chance I'd just have to tear it out again later.

gnzlbg:
So this took about one hour to make, I was surprised that this can be done without touching criterion itself at all: I actually initially implemented it as a separate crate that just depended on criterion.

AFAICT, every solution using custom-test-frameworks is going to need these two pieces: a custom test runner, and a procedural macro (or potentially more than one, depending on how fancy the API becomes). This PR implements these two pieces in the simplest way possible. 

Once you decide to look into custom-test-frameworks, this PR is probably a good start for that.

PR #212, opened 2018-09-22
Fix missing output in iter_with_large_setup

### LINKED ISSUE #211: `iter_with_large_setup` is missing output
Unlike `iter_with_setup`, `iter_with_large_setup` doesn't have output type, and so accepts only functions that don't return anything, and in addition to not having `black_box` means that it can have `routine` completely optimised out.

RReverser:
Fixes #211: `iter_with_large_setup` not taking output from the routine, effectively lettting compiler to completely optimise it out.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -453,17 +453,17 @@ impl Bencher {
     /// criterion_main!(benches);
     /// ```
     #[inline(never)]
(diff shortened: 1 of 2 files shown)


bheisler:
Huh. I'd never noticed that. Good catch, and thanks for the pull request!

I suppose it's technically a breaking change. It's very unlikely to impact anyone, the current code is so broken without the change, and the fix is so small that I'm inclined to just accept it anyway.

PR #214, opened 2018-09-25
Fix iter_with_* overhead

### LINKED ISSUE #210: `iter_with_large_drop` is much slower than `iter`
On simple benchmarks which might have potentially expensive `Drop` I've noticed up to 20% slowdown from using `iter_with_large_drop` instead of just `iter`, whereas normally I'd expect it to be always the same or faster. I suspect the reason is the extra moves and `.push` in the iteration loop which skew the measurements.

RReverser:
- Enhance iter_with_* benchmarks  …
  - Make sure that allocation is not optimised to `calloc` (which Rust does for common zero-fill allocations).
  - Measure same action across these benchmarks with same benchmark configs for easier comparisons.
  - Rename benchmarks to corresponding methods for clarity.
  - Add counterparts with unnecessary iter_with_large_* calls to measure overhead of these functions.
- Reduce overhead from iter_with_* functions …
  - Remove `black_box` from the hot path where possible (it's necessary to prevent optimisations, but doesn't have to be part of the measurement, especially the expensive `black_box` fallback used on stable Rust).
  - Use `repeat_with` + `.collect` in `iter_with_large_drop` to avoid `.push` which has to check capacity etc. on each iteration, whereas `repeat_with` lets `Vec` allocate everything in place in a tight loop.

For `iter_with_large_drop` and `iter_with_large_setup` with not-really-large drops this brings 20% and 15% increase in performance correspondingly, bringing them on par with performance of regular `.iter`.

Fixes #210.

### DIFF
--- a/benches/benchmarks/iter_with_large_drop.rs
+++ b/benches/benchmarks/iter_with_large_drop.rs
@@ -5,14 +5,31 @@ use std::time::Duration;
 
 const SIZE: usize = 1024 * 1024;
 
-fn alloc(c: &mut Criterion) {
+fn large_drop(c: &mut Criterion) {
     c.bench(
-        "alloc",
-        Benchmark::new("alloc", |b| {
-            b.iter_with_large_drop(|| (0..SIZE).map(|_| 0u8).collect::<Vec<_>>())
-        }).warm_up_time(Duration::new(1, 0))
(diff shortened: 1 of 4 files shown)


bheisler:

PR #222, opened 2018-10-09
Add deps.rs shield

lu-zero:
Make easier to spot if a dependency is outdated.


bheisler:
Hey, thanks for the suggestion.

I'm generally OK with leaving outdated dependencies as-is unless there's a reason to do otherwise. Criterion.rs isn't security sensitive and I've never yet found a bug that I traced to upstream code, so there isn't much need to keep up to date with the latest versions of dependencies - the worst that happens is users need to wait a few extra seconds compiling some older libraries the first time they run a benchmark.

PR #226, opened 2018-10-11
Fixes link to example html report

joelgallant:
See [here](https://japaric.github.io/criterion.rs/book/user_guide/html_report.html) - the relative link doesn't need `user_guide/`


bheisler:
Whoops. Thanks!

PR #227, opened 2018-10-12
When gnuplot --version errors, don't panic when parsing an empty stdout

TooManyBees:
There's an open `FIXME: Parsing may fail` comment in `plot::version()` which this PR partially addresses.

In the panic that I encountered, `gnuplot --version` failed completely (on OSX, it was linked to a version of a dylib that was no longer present on the system). The command printed its error message to stderr, leaving nothing in stdout for the function to parse.

The main issue that prompted this PR was that the panic on unwrapping a `None` gave no indication of what caused it, and I needed to search through the source on GitHub to find enough clues to work around.

This PR leaves the rest of the `unwrap`s untouched, but I could change those into descriptive `expect`s or wrap them in a `try!` like the intial command is, depending on what the intended behavior is.

### DIFF
--- a/plot/Cargo.toml
+++ b/plot/Cargo.toml
@@ -14,6 +14,8 @@ license = "MIT/Apache-2.0"
 byteorder = "1"
 cast = "0.2"
 itertools = "0.7"
+failure = "0.1.2"
+failure_derive = "0.1.1"
 
 [dev-dependencies]
 itertools-num = "0.1"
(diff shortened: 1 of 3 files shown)


TooManyBees:
rad, I've never had a reason to use Failure yet, I'll check it out

bheisler:
Looks good to me. Thanks!

PR #228, opened 2018-10-14
Only log warnings about missing gnuplot unless explicitely requested

KillTheMule:
In which case we panic instead of continuing silently.

So this was bugging me a bit, always getting the messages about missing gnuplot. I suggest only logging a message in the `default()` constructor.

During fixing that I saw that `with_plots` silently continues if gnuplot isn't found. In my book that's not a good thing, so I changed it to panic with a message (other setting functions also panic if you use a wrong parameter, so that seemed the way forward). Note that this does what I think it should if one uses it after `Criterion::default()`, but if someone put up a custom `Criterion` struct (it's `pub` after all), there might be some disconnect. I suggest redoing the `criterion_plot::version` within `with_plots`, but I'll wait on your judgement on that.

(e) I'm not sure what's borking the formatting here, looks ok locally, and I did not use any tabs.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -535,8 +535,8 @@ impl Default for Criterion {
             plotting = if criterion_plot::version().is_ok() {
                 Plotting::Enabled
             } else {
-                println!("Gnuplot not found, disabling plotting");
-
+                info!("Gnuplot not found while creating default \
+                       settings, disabling!");
                 Plotting::NotAvailable
             };
             reports.push(Box::new(Html::new()));
@@ -688,7 +688,8 @@ impl Criterion {
     /// Enables plotting
     pub fn with_plots(mut self) -> Criterion {
         match self.plotting {
-            Plotting::NotAvailable => {}
+            Plotting::NotAvailable => panic!("Plotting explicitely requested \
+              but gnuplot was not found!"),
             _ => self.plotting = Plotting::Enabled,
         }
 


KillTheMule:
As far as I can see, there are 3 main ways to configure plotting (not meaning actively, just meaning that the value of the setting has to be determined):

1. Just use `Criterion::default()` to get a `Criterion` struct that has `plotting` set appropriatedly.
2. Do 1. and then 
2a. Use `.with_plots`
2b. Use `.without_plots`
3. Do 1. and then use `.configure_from_args` via the `--noplot` argument.

Did I overlook something? They could of course be combined, but that wouldn't make a lot of sense to me. Seems we have to call `default` in any case to get a `Criterion` struct. That means, in any of the 3 cases we get a warning message if we do not have gnuplot installed.

Let me try to summarize behavior:

1. If `feature(html_reports)` is disabled, set to `NotAvailable`. Otherwise, check for it, set to `NotAvailable` or `Disabled` appropriately, print out a warning if `NotAvailable`.
2aa. If `NotAvailable` from 1., do nothing.
2ab. If `Disabled`, set to `Enabled` (no check).
2ba. If `NotAvailable`, do nothing.
2bb. If `Enabled` or `Disabled`, set to `Disabled`.
3a. If `NotAvailable`, do nothing.
3b. If `Enabled` or `Disabled`, set to disabled.

So here are the things I'd suggest to improve:

1. Everything has to go through this, so printing a warning on `NotAvailable` is noisy. Appropriate (maybe) for use case 1, but not for 2. and 3.
2aa. Don't run if `NotAvailable`, but explicitely requested (i.e. panic).
2ab. Should check before setting to `Enabled`, do not run if not available (i.e. panic)

Does that make sense to you? On second thought, I'd mainly drop loggin here, because we're trying to communicate a mismatch between standard behavior and user request, which should be explicit rather than just in the log. Might still log it, though...

So 2aa and 2ab should be easy to implement just the way I did, are you ok with that?

1. is a bit harder. One idea would be to make users of 2 or 3 not go through `default` but provide a different constructor for these cases. Could be as easy as copying `default()` and then remove the verbosity, calling it `new_silent` or whatnot. We could make `default()` use that to remove code duplication. We could also make `new`, and have it return a `CriterionUnfinished` struct that needs to be used with a `plots(self, enable: bool) -> Criterion` method to finalize (or more methods with less args). This could get out of hand if you introduce more config options that need a similar treatment..

Thoughts?

Side note: I found a small snag, and have a bit of time on my hands, so I felt like contributing to make this lib a bit better. If that's all too much overhead for such little gain to you, feel free to tell me to back off. I can certainly live with the current behavior and I'll keep using criterion :)

bheisler:
Hey, sorry I've been so slow getting back to you on this, I haven't had much time for Criterion.rs work lately.

Currently `Criterion::default` sets plotting to either `NotAvailable` or `Enabled` (not `Disabled`), and I want to keep it that way. It's uncommon to set the plotting option either way, so changing the default would be a major change in behavior for most users. Also, some of your ideas would require a breaking change to users, so I'd prefer to avoid that.

However, I agree that when the user does disable plotting, there's no need to warn about gnuplot not being available. Also, if the user explicitly enables plotting, we should fail outright if plotting cannot be enabled.

The `Plotting` enum is not exposed to the user, though, so we can add a new value to that. Here's what I'm thinking:

- Add a `Unset` variant to `Plotting`, which is the default.
- `with_plots` checks if gnuplot is available and panics if it isn't, otherwise sets the value to `Enabled`
- `without_plots` sets the value to `Disabled`
- If the value is `Disabled`, then `configure_from_args` doesn't need to do anything. Otherwise, it should check for `--noplot` and set the value accordingly, printing a message (but not panicking) if `--noplot` is not present and gnuplot cannot be found.

bheisler:
Thanks for the pull request, but I've decided to implement this myself.

PR #230, opened 2018-10-14
Remove drop from timed loop in `iter_with_large_setup`

chrisduerr:
I found the behavior of the `iter_with_setup` method very useful
because it allows returning values with significant drop times
to exclude it from the benchmark.

The `iter_with_large_setup` method currently does not follow the
same procedure as the normal `iter_with_setup` method, so this
is a proposal to streamline the two methods and make it possible
to drop things outside of the timed scope.


bheisler:

PR #232, opened 2018-10-24
Set handlebars version range

sunng87:
Keeping handlebars version in 1.0.x range for rust 1.23.0 support. I'm going to merge Pest 2.0 in Handlebars 1.1.0, which doesn't work with Rust 1.23.


bheisler:
Thanks for the pull request!

PR #236, opened 2018-11-13
Update the dependencies

lu-zero:
It is mainly needed for the `failure` dep, I will send another patch later to update `rand` since it requires code changes.

Currently criterion tests pass on stable but do not build on nightly even before this update.


bheisler:
I've updated the dependencies. Thanks for the pull request anyway, though!

PR #238, opened 2018-11-17
Simplify Gaussian kernel

KyleSiefring:
Most of the time is still spent calculating the exponent but this improves performance a little. I would give benchmark figures but the numbers aren't exactly consistent on my laptop.

### DIFF
--- a/stats/src/univariate/kde/kernel.rs
+++ b/stats/src/univariate/kde/kernel.rs
@@ -22,7 +22,8 @@ where
     fn evaluate(&self, x: A) -> A {
         use std::f32::consts::PI;
 
-        (x.powi(2).exp() * A::cast(2. * PI)).sqrt().recip()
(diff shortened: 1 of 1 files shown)


bheisler:
Hey, thanks for trying Criterion.rs and thanks for the pull request! Thanks for your patience as well.

Could you give me an idea of the reason for this change? This isn't in a hot code path (as far as I know, anyway).

KyleSiefring:
The code I'm working with has some very small functions that I want to benchmark and the benchmarks can be quite slow. Most of the time seems to be spent in this kernel.

KyleSiefring:
https://github.com/xiph/rav1e
cargo bench get_sad

bheisler:
Thanks for the contribution!

Unfortunately, after some more-intense benchmarking it looks like this doesn't actually improve performance.

I've updated the benchmarks to reduce the instability (I had dialed some of the settings down to make the benchmarks run in CI, but now that I don't run the benchmarks in CI anymore that no longer makes sense).

Issue #241, opened 2018-11-30
Show changes in throughput


ggriffiniii:
When defining a benchmark with a throughput and comparing runs using --baseline I get results like the following:
```
bench_small_input/decode_slice/50                                                                             
                        time:   [26.444 ns 26.454 ns 26.466 ns]
                        thrpt:  [1.7595 GiB/s 1.7602 GiB/s 1.7609 GiB/s]
                        change: [-26.246% -26.182% -26.119%] (p = 0.00 < 0.05)
                        Performance has improved.
```

It displays the time of the current iterations, throughput of current iterations, and how the time compares to the baseline. This is tremendously useful, but it would be nice to show how the throughput changed compared to the baseline. You can derive the throughput from the above, but it's not the most intuitive. Some people that are thinking about their benchmark results in terms of throughput may even see the change of 26% and assume the throughput went up by 26% when in fact it went up by 35%.

I'm open to trying to implement the change and sending a PR, but I would like to have a consensus before I start. If you agree this would be worthwhile, do you have an opinion on how the text output above should look?

maybe
```
bench_small_input/decode_slice/50                                                                             
                        time:   [26.444 ns 26.454 ns 26.466 ns]
                        thrpt:  [1.7595 GiB/s 1.7602 GiB/s 1.7609 GiB/s]
                 change:
                        time:  [-26.246% -26.182% -26.119%] (p = 0.00 < 0.05)
                        thrpt: [+35.xxx% +35.xxx% +35.xxx%] (p - 0.00 < 0.05)
                        Performance has improved.
```

It probably makes sense to also add the information to the html reports. Again, I would be happy to take a stab at it if you have an idea how you would like the data presented.

ggriffiniii:
Just to have something concrete. I threw together a quick commit that does what I want and mostly punts on how best to display the data.

https://github.com/ggriffiniii/criterion.rs/commit/a0326ea4a75e234b6c350e3f23a22c09a76a12fa

It currently just reports:
```
bench_small_input/decode_slice/50                                                                             
                        time:   [26.754 ns 26.768 ns 26.784 ns]
                        thrpt:  [1.7386 GiB/s 1.7396 GiB/s 1.7406 GiB/s]
                        change: [-25.410% -25.343% -25.280%] (p = 0.00 < 0.05)
                        thrpt:  [+33.833% +33.946% +34.067%] (p = 0.00 < 0.05)
                        Performance has improved.
```

bheisler:
[...]

PR #242, opened 2018-12-06
Display throughput changes in reports

ggriffiniii:
Display throughput changes in reports as described in #241


bheisler:
Thanks for the pull request! It might take me a while to get to it. Don't worry about the Travis failure just yet, I need to fix up the Travis builds.

PR #257, opened 2019-01-15
 Replace `unsafe` thread-scoped and num_cpus with rayon 

huonw:
This does not seem to result in any regressions in performance, and
possibly some small (~5%) speed-ups. This is technically a breaking
change, as it required adding some trait bounds to some functions, but
anyone that is actually broken risks having memory
unsafety (non-`Send` types being sent across threads).

Work towards #195.

Also, since rayon makes it so easy:

Use rayon's par_sort_unstable_by for dramatic `percentile` speedup 

For the benchmarks that call `percentile` on large arrays (e.g. 1
million in the ones below), this gives a very large speed-up: >80%
less time taken/5x faster.

```
stat_f64_iqr            time:   [16.732 ms 17.470 ms 18.287 ms]
                        change: [-81.695% -81.052% -80.361%] (p = 0.00 < 0.05)
stat_f64_median         time:   [13.657 ms 13.809 ms 13.953 ms]
                        change: [-85.984% -85.696% -85.421%] (p = 0.00 < 0.05)
```

(It means that it's a `cargo bench` run could finish on my machine. I wasn't patient enough to leave my computer unused for benchmarks like `median_abs_dev` to finish before: the prediction was >10 minutes for just that one.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Fixed
+
+- Criterion.rs now uses `rayon` internally instead of manual `unsafe` code built on top thread-scoped.
+
 ## [0.2.6] - 2018-12-27
 ### Added
 - Added `--list` command line option, which lists the benchmarks but does not run them, to match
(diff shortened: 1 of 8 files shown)


bheisler:
Wow! That is awesome, thank you!

I'm not concerned about the breaking change; I'm going to merge criterion-stats into criterion in 0.2.8 anyway, so that code will no longer be "public API" (not that it ever was, really).

Issue #267, opened 2019-02-16
Add a colored output on Windows?


alekseysidorov:
At the moment criterion output on windows doesn't support colors, thus it is harder to understand benchmark results that on *nix.

bheisler:
Hey, thanks for the suggestion! That seems like a reasonable feature to add, but I don't typically develop on Windows. Pull requests would be welcome, though!

vpetrigo:
Hey @alekseysidorov,

Would you please specify which Windows version you mean? I'm on Windows 10 (build 1809 - 10.0.17763) and color output works fine for me.

So that's is the short snippet I added into the `report.rs`:

```rust
    #[test]
    fn test_windows_colors() {
        let cli = CliReport::new(false, true, false);
        let hello = "Hello";

        println!("{}", cli.green(hello.to_string()));
        println!("{}", cli.red(hello.to_string()));
    }
```

And how it looks in the plain `cmd`:

![Hello Test](https://i.ibb.co/GxzhzZ0/image.png)

That's how the benchmarks output of `criterion.rs` itself looks like in my `cmd`:

![Criterion Bench](https://i.ibb.co/tXSR9Lf/image.png)

I may elaborate on that, but need more info about the system you experienced such an issue. 😄

vpetrigo:
Small update on that: seems like that issue with lack of colors in Windows terminals has been fixed in rust-lang/cargo#4722. So I was able to reproduce OP's issue with Rust 1.26.0, but it went away after I switched back to the `stable-msvc` toolchain.

alekseysidorov:
I am using rustc version `1.32.0` and `powershell` as shell. Sometimes I run tests in `git-bash` but I never tried to run criterion benchmarks in `cmd.exe`.

vpetrigo:
@alekseysidorov try to run you benchmarks with the following command: `cargo bench -- --color=always`. I guess there are some issues with traking PowerShell color output features.

alekseysidorov:
With additional argument works fine in powershell
![2019-03-01_12-14-14](https://user-images.githubusercontent.com/83360/53628343-deb99a80-3c1b-11e9-82f5-8c636be3f9ff.png)
But without them colors do not work:
`cmd.exe`
![2019-03-01_09-58-03](https://user-images.githubusercontent.com/83360/53628403-01e44a00-3c1c-11e9-91c4-14bd59e54ed5.png)
`powershell`
[...]

PR #271, opened 2019-03-05
Enable proper text coloring on Windows

vpetrigo:
That PR addresses #267. According to rust-lang/cargo#4722 the misbehaving of color output with `rustc` and `cargo` has been fixed. Also `atty` is able to properly determine whether we use TTY on Windows or not. That small fix correctly enables color ouput while using Windows `cmd` and `powershell` utilities and also does not break down compatibility with Unix terminal programs.

I'd like to ask @alekseysidorov to validate whether that fix properly works on his machine. My test configuration was the following:
* Windows 10 (`build 1809 - 10.0.17763`)
* `rust 1.32.0`
* `rust 1.33.0`

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -972,7 +972,7 @@ scripts alongside the generated plots.
                 enable_text_coloring = false;
                 enable_text_overwrite = false;
             }
-            _ => enable_text_coloring = cfg!(unix) && stdout_isatty,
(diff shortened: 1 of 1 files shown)


alekseysidorov:
Yes with this PR colored output works fine.

bheisler:
Then this looks good to me. Thanks for the pull request!

PR #277, opened 2019-03-09
Issue 272

pierrechevalier83:
Most of the code to avoid duplicate directories was already in place.
The only problem was that the cache of directories that is checked for duplicates
didn't outlive a single call to `Benchmark::run.`
Fix it by making this cache a member of `Criterion` so it does live long enough to
be useful in the situation that my usecase exposed.


bheisler:
Thanks for the pull request. Unfortunately, due to some IRL stuff I probably won't be able to review it for a while.

bheisler:
Yikes. Thanks for your patience. This looks good to me.

PR #282, opened 2019-04-03
Fix trailing spaces in filenames on Windows

RReverser:
See the comment - basically the problem, is that Windows doesn't allow trailing spaces in filename, and instead silently trims them.

This works more or less well for file access, since the filename is trimmed in the same way both on creating and on access, but when used with directories, trying to create something like `dir /file` will simply fail with no apparent reason.

This was a problem before, but it was additionally revealed by the introduction of MAX_DIRECTORY_NAME_LEN since now name can be trimmed to a one with a trailing space much more frequently.

### DIFF
--- a/src/report.rs
+++ b/src/report.rs
@@ -83,13 +83,27 @@ pub fn make_filename_safe(string: &str) -> String {
         "_",
     );
 
-    // On Windows, file names are not case-sensitive, so lowercase everything.
-    if cfg!(target_os = "windows") {
-        string = string.to_lowercase();
-    }
-
     // Truncate to last character boundary before max length...
     truncate_to_character_boundary(&mut string, MAX_DIRECTORY_NAME_LEN);
+
+    if cfg!(target_os = "windows") {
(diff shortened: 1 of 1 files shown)


RReverser:
@bheisler I see CI failed due to usage of `.trim_end`.

It used to be called `.trim_right`, and I can change to that for older Rust support, but on another hand it's soft-deprecated (with a warning) in newer Rust versions, so there's a tradeoff.

What would you prefer?

bheisler:
Yes, I've run into the same problem in TinyTemplate. I'd say use `trim_right` for backwards-compatibility, and suppress the deprecation warning. Maybe add a TODO comment about updating once we no longer require 1.30.0 or later.

RReverser:
Ok, done.

RReverser:
Clippy is wrong here though :/

RReverser:
Fixed the Clippy thing (hopefully).

bheisler:

PR #283, opened 2019-04-03
Fix overhead of iter_batched{_ref}

RReverser:
Looks like newly added `iter_batched` and `iter_batched_ref` methods re-introduced same problem as I've fixed in the past for `iter_with_*` methods in #214.

This essentially applies same fixes to the new methods.

Original quote from #214:
> Remove `black_box` from the hot path where possible (it's necessary to prevent optimisations, but doesn't have to be part of the measurement, especially the expensive `black_box` fallback used on stable Rust).

Some numbers from `cargo bench -- overhead/iter_batched`:

| name | diff |
|-|-|
| `iter_batched_small_input` | -22% |
| `iter_batched_large_input` | -54% |
| `iter_batched_per_iteration` | no change |
| `iter_batched_ref_small_input` | -26% |
| `iter_batched_ref_large_input` | -87% |
| `iter_batched_ref_large_input` | no change |

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -475,14 +475,14 @@ impl Bencher {
             while iteration_counter < self.iters {
                 let batch_size = ::std::cmp::min(batch_size, self.iters - iteration_counter);
 
-                let inputs = (0..batch_size).map(|_| setup()).collect::<Vec<_>>();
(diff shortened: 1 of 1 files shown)


bheisler:
Ah, right. I'd completely forgotten about that. Good catch, thanks!

Issue #261, opened 2019-01-25
Roadmap for 0.3.0


bheisler:
I'm planning to take some time off working on Criterion.rs shortly, but before I do I'd like to describe my plans for 0.3.0 and request some comments from the community. 0.3.0 is the first breaking-change release since Criterion.rs became stable-compatible, and I have a handful of things I'd like to change.

### Removing Benchmark/ParameterizedBenchmark

I initially created these as a generalization of the various `Criterion::bench_*` functions and they mostly work, but there are a few flaws in this design. For instance, there isn't an obvious way to handle cases like #225, where the user wants to construct a large input once. Additionally, it imposes strange limitations on the parameter types (why must they implement Debug?) and strange contortions to deal with lifetimes (#260). Additionally, sometimes the user doesn't want to have the debug representation of the parameter be used as the parameter in the benchmark ID - for example, if I'm testing a parser I don't want my benchmark to be called "my_group/my_function/Some really really long file full of text that my parser parses but isn't really relevant to the benchmark...". After thinking about this for a while, I wondered why we even need these structures at all? After all, they basically just define a nested for loop:

```rust
for function in self.functions {
    for parameter in self.parameters {
        // Run a benchmark of this function with this parameter
    }
}
```

I'm sure the user is capable of writing their own for-loops, so why not let the user provide that? Here's a sketch of what I'm thinking:

```rust
let criterion: &mut Criterion = ...;

// For simple benchmarks, the existing bench_function would still work.
criterion.bench_function("my_function", move |b| {
    b.iter(|| my_function())
}):

// Benchmark and ParameterizedBenchmark, as well as bench_functions and 
// bench_function_over_inputs would be replaced with a new BenchmarkGroup struct
let mut benchmark_group = criterion.benchmark_group("Hashes");

// The user can compare different functions by just calling bench_function multiple times
// This would produce benchmarks called "Hashes/SHA1" and "Hashes/MD5".
benchmark_group.bench_function("SHA1", |b| b.iter(|| sha1sum("some input string"));
benchmark_group.bench_function("MD5", |b| b.iter(|| md5sum("some input string"));

// `finish` would consume the benchmark group structure and generate the summary reports.
// The benchmark group could be marked `must_use` to warn people if they forget?
benchmark_group.finish();

// Handling multiple parameters is straightforward
let mut benchmark_group = criterion.benchmark_group("SizedHashes");
for input_size in vec![10, 100, 1000, 10000, 100000] {
[...]

PR #286, opened 2019-04-12
WIP: Linux perf counters

Licenser:
Heya,
I am looking at including linux perf counters (such as cpu cycle, cach misses etc.) to criterion. It'd allow not only drive benchmarks on pure time but also other vectors that affect performance.

At this point this is nothing more then a proof of concept how hard that is in rust turns out it's a bit of legwork but not too hard.

Before going on I want to make sure there is interest in merging something like this. Since it depends on nightly code it'd need to be fenced of behind a feature flag but it I see it being possibly quite valuable non the less.

If ther is interest I'd love to get some input how to best put it in criterion itself so it doesn't get in the way of design or future goals.

Chears!

my TODO (Before WIP end so I don't forget things):
- [ ] Fence of unstable bits
- [ ] Remove rust-toolchain

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,6 +33,7 @@ rayon = "1.0.3"
 # These are needed for the minimal-versions build
 rayon-core = ">=1.4.1"
 libc = ">=0.1.5"
+perfcnt = "0.4"
 
 [dev-dependencies]
 tempdir = "0.3.7"
--- a/rust-toolchain
+++ b/rust-toolchain
@@ -1,0 +1,1 @@
+nightly
--- a/src/analysis/mod.rs
+++ b/src/analysis/mod.rs
@@ -74,8 +74,8 @@ pub(crate) fn common<T>(
         return;
     }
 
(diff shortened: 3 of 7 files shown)


bheisler:

I would like to provide access to things like this, certainly. At the same time, I'm not sure that this is the best way to do it. For one thing, recording the performance counters will affect and/or be affected by recording the wall-clock time, and it would be nice if people who only care about one or the other didn't have to pay that cost. The other thing is that I'd rather not have to re-work the guts of Criterion.rs every time someone wants a new performance measurement. In particular, with this approach we'd have to have a lot of awkward conditional-compilation right in the core of Criterion's measurements which strikes me as a risky thing to do - tricky to ensure that the different conditionally-compiled variants of things stay in sync.

I have some rough ideas in #261 for ways to support adding additional measurements as an extension so that benchmarks could specify exactly what they want to measure. That would make it possible to add measurements using an external crate. That said, I don't know when I'll get around to actually building that.

Licenser:
👍 #261 looks very relevant I put in some notes :)

Issue #265, opened 2019-02-02
Benchmark scalability for multiple cores?


jbreitbart:
I am wondering what is the best approach to do not only check the runtime of a function, but also check its scalability when used by multiple threads. For example, it may be interesting to check the performance of having multiple threads add an element to a hashmap or make sure that my latests changes do not break the scalability of an algorithm.

Is using `bench_function_over_inputs` and pass the number of threads as a parameter the correct approach?

bheisler:
Hey, thanks for trying Criterion.rs!

Passing the number of threads as a parameter is one reasonable way to do it. It does mean you can't (easily) pass other things as the parameter, though. I would recommend using multiple "functions" where each function is the same function with a different number of threads.

Unfortunately, I think you'll have to manually iterate your code inside each thread in order to create thread contention. That will make the measurements harder for Criterion.rs to analyze and harder for the reader to understand, which is not great.

I think your use case calls for a `bencher.iter_custom(|iters: usize| {...} -> Duration)` function. If that were available, you could spin up your threads and signal them to each run `iters` iterations of the hashmap insertion or whatever. You would have to write your own timing measurements (by which I mean use `Instant::now()` and `Instant::elapsed()`) but it would get you much better measurements.

jbreitbart:
Thanks for the quick reply (and sorry for me being slow, I don't have much time to work on my rust stuff). I general, I think it would be great if there were explicit support for multi-threaded measurements and analysis. Here is some feedback / questions on your suggestions.

Having one benchmark functions per thread count makes it hard to benchmark on machines with different number of cores, which is why I have moved forward with the idea of passing the thread count as a parameter. It is not great, especially as I am also benchmarking thread creation in my current code.

I am not sure how it would look like if there would be a `bencher.iter_custom(|iters: usize| {...} -> Duration)` function. Would I just do the time measurements on my own or would I also have to do the analysis? Doing the measurements seems to be fairly easy, but I think doing the analysis should be in a library as one can easily make a mistake there. There are also quite a lot of analysis that could make sense depending on test case.

Would support for rayon be easier than support for multiple threads in general?

bheisler:
> I am not sure how it would look like if there would be a `bencher.iter_custom(|iters: usize| {...} -> Duration)` function. Would I just do the time measurements on my own or would I also have to do the analysis?

Your code would simply do the measurement; Criterion.rs will take care of the analysis.

> Would support for rayon be easier than support for multiple threads in general?

[...]

PR #307, opened 2019-07-05
Provide iter_custom

gz:
Took a stab at handling the issue https://github.com/bheisler/criterion.rs/issues/265.
Provides a vary basic implementation of iter_custom, let me know if this is what you had in mind as an  interface.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -347,7 +347,40 @@ impl Bencher {
         self.elapsed = start.elapsed();
     }
 
-    /// iter_custom as discussed in https://github.com/bheisler/criterion.rs/issues/265
+    /// Times a `routine` by executing it many times and relying on `routine` to measure it's own execution time.
+    ///
+    /// Prefer this timing loop in cases where `routine` has to do it's own measurements to
+    /// get accurate timing information (for example in multi-threaded scnearios where you spawn
+    /// and coordinate with multiple threads).
+    /// 
+    /// # Timing model
+    /// Can be customized and depends on the implementation of `routine`.
+    ///
+    /// # Example
+    /// ```rust
+    /// #[macro_use] extern crate criterion;
+    /// use std::time::Instant;
+    /// 
+    /// fn foo() {
+    ///     // ...
+    /// }
+    ///
+    /// fn bench(c: &mut Criterion) {
+    ///     c.bench_function("iter", move |b| {
+    ///         b.iter_custom(|| {
+    ///             let start = Instant::now();
+    ///             for _i in 0..iters {
+    ///                 black_box(foo());
+    ///             }
+    ///             start.elapsed()
+    ///         })
+    ///     });
+    /// }
+    ///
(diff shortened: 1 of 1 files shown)


bheisler:
Hey, thanks for the pull request. This looks good so far (although I spotted a typo - "scnearios").

I'd like to discuss the design a bit, though. What you have here is good, but this will interact with the new custom measurements feature in 0.3.0 which allows for custom implementations of a `Measurement` trait to measure different things than just wall-clock time and to return different types than just Duration. However, we still need to have a Duration to allow Criterion.rs to perform its warmup phase for the right length of time. This leaves a few options:

- We could set up the impl's so that the lambda passed to `iter_custom` returns just a Duration, and then set up the impls so that it can only be used with measurements that also return Duration. This is obviously restrictive, but it also means that we can't really extend it to work with custom measurements without a breaking change later.
- The lambda could instead return a tuple of Duration and Measurement::Value, to allow it to work with custom measurements, but this would impose more work on people using iter_custom. It also means that measurements that do naturally return a Duration have to return the argument twice, which is kind of awkward.
- We could have the lambda just return Measurement::Value and infer the duration based on how long it takes to execute the lambda. This is how the other measurements work, and it's how I'd prefer to implement this, but there may be some use cases where the time to execute the lambda doesn't accurately represent how long it takes to call the function (eg. if you have to spin up an external process).

I'm not sure what makes sense here. What do you think?

gz:
> We could have the lambda just return Measurement::Value and infer the duration based on how long it takes to execute the lambda. This is how the other measurements work, and it's how I'd prefer to implement this, but there may be some use cases where the time to execute the lambda doesn't accurately represent how long it takes to call the function (eg. if you have to spin up an external process).

That case wouldn't fit with our use-case, I think (and the one discussed in #265). We are benchmarking a set-up which spawns multiple threads and synchronizes the timing measurement using a barrier. Which is difficult to do precisely if we just measure the lambda (set-up overhead is large). 

> The lambda could instead return a tuple of Duration and Measurement::Value, to allow it to work with custom measurements, but this would impose more work on people using iter_custom. It also means that measurements that do naturally return a Duration have to return the argument twice, which is kind of awkward.

I see, I'm not familiar with the Measurement type and how it works in 0.3 (aside: could the Duration just be embedded to be always included in the Measurement::Value and (be overridden by the lambda in iter_custom)). Other than that, this seems like an acceptable trade-off, given that the only time a user does iter_custom is when he has a pretty special use-case anyways.

bheisler:
That's fair. You're right that `iter_custom` is only needed for unusual cases, so it's not unreasonable to expect callers to do a bit more work.

Would you mind rebasing this onto the `v0.3.0-dev` branch?

bheisler:
I've merged this into v0.3.0-dev manually. Thanks for the pull request!

gz:
Thanks, sorry I was unable to respond to this due to an accident.

PR #310, opened 2019-08-05
Remove byteorder-dependency

lukaslueg:
This removes the `byteorder`-dependency, which is not needed since 1.32. This raises the minimal Rust version to 1.32, which is in line with "within last three stable versions". Dropping a dependency is worth the change IMHO.


bheisler:

I didn't even know about this until I saw your changes. I agree, it's nice to be able to drop that dependency. I'll just merge this into my 0.3.0-dev branch manually.

bheisler:
Merged into v0.3.0-dev. Thanks!

Issue #149, opened 2018-05-28
C


antifuchs:

bheisler:
Hey, thanks for trying Criterion.rs, and thanks for the suggestion.

Yeah, that's reasonable. We could probably skip adding configuration (at least for now) and assume that anyone who configures a throughput metric on their benchmarks at all is probably more interested in the throughput than the execution time.

I probably won't get around to implementing this right away, but pull requests would be welcome.

gz:
I would be willing to try and tackle this since I'd like to have this feature too. Can @bheisler maybe give me a pointer to the relevant function that I need tu modify for this?

bheisler:
I think you'd need to modify more than one function...

For now, let's scope this to just adding throughput charts to the per-benchmark reports. Reporting throughput on the summary reports raises a lot of complicated questions and edge cases (Would you want to have both throughput and execution time on the summaries? Should the violin plots show throughput instead of execution time? What if some of the benchmarks in a group have no throughput? What if they have different kinds of throughput?).

`src/html/mod.rs` is the entry point for generating the HTML reports. You probably want to edit `measurement_complete` to have it generate one or more new plots if there is a throughput (if `measurements.throughput.is_some()`). You'll need to convert the average iteration times in `measurements.avg_times` to throughput numbers and define the new plots in `src/plot`, probably in a new file. Depending on how you want to display these plots, you might also need to update the `benchmark_report.html.tt` template.

This will need to work with the custom measurements feature I've added to 0.3.0, so you'll need to build on top of the `v0.3.0-dev` branch. That will probably require a breaking change to the `ValueFormatter` trait, so it would have to be done before I release 0.3.0 or wait for 0.4.0.

Yeah, this isn't a trivial feature to add, partly because Criterion.rs' internal code isn't as clean as I'd like and partly because it interacts with some other features currently in development.

gz:
Thanks that will be very helpful, I will try to look into these pointers to see what I can come up with on the 0.3 branch.

PR #315, opened 2019-08-21
[WIP] Adding a througput chart to the benchmark summary.

gz:
A first version to address  #149. This adds a throughput chart to the summary page.

Some more thoughts:

It was a bit of a nuisance to get the Throughput struct down to the reporting and plotting, maybe there is an easier, better way that I don't know?

Also, plot::line_comparison_throughput is doing unwrap on Option<Throughput> so the caller should check that every BenchmarkId has a throughput (and that its of the same type probably). I can add that if the general direction of this pull request is deemed fine.

Maybe, instead of a line-plot a barchart will work better here?

To test I use: 
```
cargo bench -- from_elem
open target/criterion/report/index.html
```

### DIFF
--- a/src/html/mod.rs
+++ b/src/html/mod.rs
@@ -334,16 +334,26 @@ impl Report for Html {
             self.generate_plots(id, report_context, formatter, measurements)
         }
 
-        let throughput = measurements
-            .throughput
-            .as_ref()
-            .map(|thr| ConfidenceInterval {
-                lower: formatter
-                    .format_throughput(thr, slope_estimate.confidence_interval.upper_bound),
-                upper: formatter
-                    .format_throughput(thr, slope_estimate.confidence_interval.lower_bound),
-                point: formatter.format_throughput(thr, slope_estimate.point_estimate),
-            });
+        let throughput = measurements.throughput.as_ref().map(|thr| {
+            let typical_throughput = thr.per_second(slope_estimate.point_estimate);
+            ConfidenceInterval {
+                lower: formatter.format_throughput(
+                    thr,
+                    typical_throughput,
+                    slope_estimate.confidence_interval.upper_bound,
+                ),
+                upper: formatter.format_throughput(
+                    thr,
+                    typical_throughput,
+                    slope_estimate.confidence_interval.lower_bound,
+                ),
+                point: formatter.format_throughput(
+                    thr,
+                    typical_throughput,
+                    slope_estimate.point_estimate,
+                ),
+            }
+        });
 
         let context = Context {
             title: id.as_title().to_owned(),
@@ -808,7 +818,8 @@ impl Html {
             .collect();
         let mut line_tput_path = None;
 
-        let all_throughputs_match = value_types.iter().all(|x| x == &value_types[0]) && throughput_types[0].is_some();
+        let all_throughputs_match =
+            value_types.iter().all(|x| x == &value_types[0]) && throughput_types[0].is_some();
         if all_throughputs_match {
             let path = format!(
                 "{}/{}/report/lines_tput.svg",
--- a/src/html/mod.rs
+++ b/src/html/mod.rs
@@ -808,11 +808,8 @@ impl Html {
             .collect();
         let mut line_tput_path = None;
 
-        if value_types.iter().all(|x| x == &value_types[0])
-            && throughput_types[0].is_some()
-            && throughput_types.iter().all(|x| x == &throughput_types[0])
-        {
-            if let Some(value_type) = value_types[0] {
+        let all_throughputs_match = value_types.iter().all(|x| x == &value_types[0]) && throughput_types[0].is_some();
+        if all_throughputs_match {
                 let values: Vec<_> = data.iter().map(|&&(ref id, _)| id.as_number()).collect();
                 if values.iter().any(|x| x != &values[0]) {
                     let path = format!(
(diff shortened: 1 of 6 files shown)


gz:
It seems I got confused by the two structs report::BenchmarkId which has throughput and BenchmarkId which doesn't store throughput in benchmark_group.rs which I also mistakenly assumed I was passing around. 
So I can simplify this pull request quite a bit.

bheisler:
Hey, sorry I haven't had a chance to respond to this yet. Yeah, `report::BenchmarkID` is the way to go. I should also mention there's also a `ValueFormatter::scale_for_throughput` that I added specifically to support this; I haven't looked at the code changes yet so you may have seen that already.

src/plot/summary.rs
```diff
@@ -121,20 +121,116 @@ pub fn line_comparison(
...
+    for (key, group) in &all_curves.iter().group_by(|&&&(ref id, _, _)| &id.function_id) {
+        let mut tuples: Vec<_> = group
+            .map(|&&(ref id, ref tput, ref sample)| {
+                // Unwrap is fine here because it will only fail if the assumptions above are not true
+                // ie. programmer error.
+                let x = id.as_number().unwrap();
+                let mut y = [Sample::new(sample).mean()];
+                let tput = tput.clone().unwrap();
+
+                error!("{:?}", y[0]);
```
bheisler: Leftover debugging output.

bheisler (changes requested):
As you noticed, you can get the throughput out of the benchmark ID so all of the changes made to pass the throughput along are unnecessary. There's also some leftover debug output.

Otherwise this looks good. It won't make it into 0.3.0 (which I plan to release today) but it won't require a breaking change so we can add it to 0.3.1.

src/html/mod.rs
```diff
@@ -801,6 +802,39 @@ impl Html {
             }
         }
 
+        let throughput_types: Vec<_> = data
+            .iter()
+            .map(|&&(ref id, _)| id.throughput.as_ref())
+            .collect();
+        let mut line_tput_path = None;
+
+        if value_types.iter().all(|x| x == &value_types[0])
```
gz: Does this if() cover all cases?

gz:
Thanks, I updated this pull request, let's see if it passes CI.

src/html/mod.rs
bheisler: I think this test will return false if there are multiple different throughput values. For example, some benchmarks could have a throughput of `Bytes(1000u64)` and others could be `Bytes(500u64)`, and that would register as different throughput types. I think what you want is to check if they're all either `Bytes(_)` or `Elems(_)`. In that case, I think something like this should work:

```
let all_throughputs_match = value_types.iter().all(|x| x == &value_types[0]) && throughput_types[0].is_some();
if all_throughputs_match { ... }
```

You don't need the third clause because if all of the value-types match then the throughputs must all have the same variant. Maybe add a comment explaining that though, it's not obvious from reading the code.

src/plot/summary.rs
```diff
@@ -121,6 +121,106 @@ pub fn line_comparison(
     f.set(Output(path)).draw().unwrap()
 }
 
+#[cfg_attr(feature = "cargo-clippy", allow(clippy::explicit_counter_loop))]
+pub fn line_comparison_throughput(
+    formatter: &dyn ValueFormatter,
+    title: &str,
+    all_curves: &[&(&BenchmarkId, Vec<f64>)],
+    path: &str,
+    value_type: ValueType,
```
bheisler: Taking the value type as a parameter here is unnecessary; you can get that information from the throughputs in the BenchmarkIDs. Removing it also allows you to simplify the calling code a bit.

src/plot/summary.rs
```diff
@@ -121,6 +121,106 @@ pub fn line_comparison(
...
+                .set(Order::SampleText)
+                .set(Position::Outside(Vertical::Top, Horizontal::Right))
+        })
+        .set(Title(format!("{}: Comparison", escape_underscores(title))))
+        .configure(Axis::BottomX, |a| {
+            a.set(Label(format!("Input{}", input_suffix)))
+                .set(axis_scale.to_gnuplot())
+        });
+
+    let mut i = 0;
```
bheisler: Move this down to the loop where it's used.

bheisler (changes requested):
Better! Just some minor things to clean up.

gz:
Finally got around to addressing the reviews.

The build currently fails due to unrelated issues,should be fixed with #336

gz:
I changed the plotting for throughput slightly to take into account parameterized benchmarks.

Ekleog:
Hey! Just wanted to ask if there's any news on this PR, as it'd be very nice :)

gz:
Hey, the branch/PR works/worked for me -- feel free to give it a try ... Unfortunately, I couldn't quite fit my requirements for a multi-core benchmark harness into criterion (even with `iter_custom`) so I am currently not actively relying on the patch but maybe @bheisler could still check if it's something for master if it's useful for you.

h33p:
It would be really nice to have this in. Although it fails to generate a plot for logarithmic scale (min yrange can't be zero). That should can be solved by removing `set(Range::Limits(0, max_formatted[0])` (line 181 from summary.rs). If this is still of interest to @bheisler I could take over this and finish implementing the feature.

bheisler:
Hey folks, thanks for your patience. Yeah, if someone else wants to take over and finish this off, that'd be great!

lemmih:
> It would be really nice to have this in. Although it fails to generate a plot for logarithmic scale (min yrange can't be zero). That should can be solved by removing `set(Range::Limits(0, max_formatted[0])` (line 181 from summary.rs). If this is still of interest to @bheisler I could take over this and finish implementing the feature.

I'm ready to review if you take over this PR.

lemmih:
Closing due to inactivity.

PR #321, opened 2019-08-26
fix colon→semi-colon typo in book

tgpfeiffer:


bheisler:
Ah, my mistake. Thanks for the fix!

PR #330, opened 2019-09-15
Bump rand_xoshiro to 0.4

eclipseo:


bheisler:
Hi! Thanks for the pull request (and the patience) but I've decided to just remove the runtime dependency on `rand` altogether.

PR #329, opened 2019-09-13
Use OsRng from rand_core instead of deprecated rand_os

eclipseo:
Fix #328

Signed-off-by: Robert-André Mauchin <zebob.m@gmail.com>

PR #336, opened 2019-09-22
remove deprecated rand_os and added CyclesPerByte measurement

WildCryptoFox:
Includes #329. I can rebase after it is accepted separately.

The `CyclesPerByte` measurement implementation uses an x86/x86_64 specific intrinsic, so it deserves its own feature in `Cargo.toml` `cycles`. It doesn't ask `cpuid` for support but it'd panic anyway.

Using `CyclesPerByte` displays the `cycles` on all benchmarks, `cpb` for byte-wise throughput benchmarks, and `c/e` for element-wise throughput benchmarks. It would be nicer to measure both `CyclesPerByte` and `WallTime`; should they be measured independently as to not count each-other?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,9 +27,8 @@ walkdir = "2.2"
 tinytemplate = "1.0"
 cast = "0.2"
 num-traits = "0.2"
-rand_os = "0.2"
 rand_xoshiro = "0.3"
-rand_core = { version = "0.5", default-features = false }
+rand_core = { version = "0.5", default-features = false, features = ["getrandom"] }
 rayon = "1.1"
 
 [dev-dependencies]
@@ -46,6 +45,7 @@ maintenance = { status = "passively-maintained" }
 [features]
 real_blackbox = []
 default = []
+cycles = []
 
 [workspace]
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -36,9 +36,6 @@ extern crate approx;
 #[macro_use]
 extern crate quickcheck;
 
-#[cfg(test)]
-extern crate rand;
-
 #[macro_use]
 extern crate clap;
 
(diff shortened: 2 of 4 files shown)


bheisler:

Part of the reason I built `Measurement` as a trait was so that alternative measurements like this could be developed as third-party crates. Is there some particular reason you'd like this to be included into Criterion.rs itself?

Regarding the rand changes, that should probably be a separate PR but I'm strongly leaning towards dropping the rand crates entirely and switching to `oorandom` instead.

WildCryptoFox:
A separate crate is fine, though I'd expect it to be popular for at least crypto crates. At least crates.io can tell me who uses it.

Bah, this is awkward. It appears someone has already uploaded my code as a crate. https://crates.io/crates/criterion-cycles-per-byte

@YangKeao You're breaking copyright by claiming authorship. Could you please [transfer ownership](https://doc.rust-lang.org/cargo/reference/publishing.html#cargo-owner) of `criterion-cycles-per-byte` to me?

YangKeao:
@james-darkfox 

Great! I will do it right now. I will remove myself from owners right after you accept the invitation.

Beside a transfer request of github repo has been sent (If you won't accept it please tell me, I will delete my repo).  

Sorry for bothering you. I uploaded this because I need this measurement before this PR will be merged.

WildCryptoFox:
@YangKeao Accepted and removed you as an owner. Thank you. For future reference, if you need code that isn't published, embed it in your project and **reference the source with appropriate licenses followed**.

You may delete your repository. I've published the crate over on [GitLab](https://gitlab.com/sio4/code/criterion-cycles-per-byte) and on crates.io as version 0.1.1.

bheisler:
Thanks for publishing the criterion-cycles-per-byte crate; I've added a link to it from the README.

PR #337, opened 2019-09-24
Use cargo_metadata to detect target directory

### LINKED ISSUE #192: Reports are put into incorrect directory when used in a workspace
Say, I have a workspace with crates `crate-a` and `crate-b`. By default, Cargo will crate top-level directory `target` where it will put all build artifacts for these crates and their dependency.

However, if `crate-a` uses Criterion, it creates its own directory `crate-a/target` instead of using the directory provided by Cargo.

dalance:
This PR uses cargo_metadata to detect target directory instead of `CARGO_TARGET_DIR`.
This will fix  #192.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,11 +31,12 @@ num-traits = { version = "0.2", default-features = false }
 oorandom = "11.1"
 rayon = "1.3"
 regex = { version = "1.3", default-features = false, features = ["std"] }
+cargo_metadata = "0.10.0"
 
 [dependencies.plotters]
 version = "^0.2.12"
 default-features = false
-features = ["svg", "area_series", "line_series"] 
+features = ["svg", "area_series", "line_series"]
 
 [dev-dependencies]
 tempfile = "3.1"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -743,10 +743,12 @@ impl Default for Criterion {
         // - ./target/criterion
         let output_directory = if let Some(value) = std::env::var_os("CRITERION_HOME") {
             PathBuf::from(value)
-        } else if let Some(value) = std::env::var_os("CARGO_TARGET_DIR") {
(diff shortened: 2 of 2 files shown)


bheisler:
Hey, thanks for the pull request. I'm super busy IRL lately so it might take a while for me to get to it.

dalance:
OK. Thanks.

ChaiTRex:
Not all filesystems use UTF-8 filenames, so I think `String` as the type for filepaths and `.to_string_lossy()` are the wrong way to go about this. Instead, filepaths throughout Criterion.rs should use something like the [`std::Path::PathBuf`](https://doc.rust-lang.org/std/path/struct.PathBuf.html) type of [`metadata.target_directory`](https://docs.rs/cargo_metadata/0.10.0/cargo_metadata/struct.Metadata.html#structfield.target_directory).

They're easy to work with, they can be cloned, and they can be used as an argument to [`std::fs::File::open(pathbuf)`](https://doc.rust-lang.org/std/fs/struct.File.html#method.open).

dalance:
@ChaiTRex Thanks! I fixed it.

dalance:
@bheisler I rebased the latest master. Could you review again?

bheisler:
OK, thanks for all your patience.

The code change looks fine, but I'm not sure about the approach of adding a new dependency. I try pretty hard to keep Criterion.rs' dependency tree and compilation time as small as I can. In this case, it seems like we're using only a very small part of the `cargo_metadata` crate. I think it would make sense to re-implement that functionality rather than use a crate. In fact, I've just implemented it that way in cargo-criterion (see https://github.com/bheisler/cargo-criterion/commit/722b5acb3cc23272d60fb68e9d9fc0321e1da2a2), and it didn't take very much code.

dalance:
Thanks!

PR #338, opened 2019-09-24
Add missing configuration to criterion-macro output

Bobo1239:
This also resulted in missing output colorization and missing plots in the HTML report.


bheisler:
Huh, how did I miss that... Good catch, thanks!

I'd suggest the `.configure_from_args()` call should be after line 28, so that if a user overrides the default configuration they still get the CLI-argument settings. That's how regular Criterion.rs benchmarks work.

Bobo1239:
Ah yes, of course. I fixed it.

bheisler:
Thanks for the pull request!

Issue #313, opened 2019-08-10
Replace gnuplot with plotters(a pure rust crate)?


Veykril:
I just stumbled upon a pure rust [plotting library](https://github.com/38/plotters) which might be of interest to this project. Since the `criterion-plot` crate is still in need of a maintainer I thought it might be an interesting idea to replace it with this crate instead. It would also remove the need of having `gnuplot` installed to generate plots.

bheisler:
Hey, thanks for the suggestion.

That's great! I didn't know about this library until now. I agree that it would be nice to replace gnuplot with a pure-Rust solution. Unfortunately I'm kinda busy with 0.3.0 and various IRL stuff right now, so this won't be a priority for me, but pull requests would be welcome. If anyone does want to implement this, I'd prefer to add this as an optional Cargo feature and maintain the two backends in parallel, at least for a while, rather than abruptly switching over.

38:
I am the author of `plotters` crate. I am currently working on the rendering performance for my crate. And `criterion` works perfect for me. But I realized on thing, `criterion` depends on the `gnuplot`. And I have a machine which doesn't have it installed.

This makes me think about using `plotters` to visualize result and I realize there's already a issue. 

So I just tried a quick prototype, the output looks like this: 

![image](https://user-images.githubusercontent.com/3023614/67240215-3f495200-f40e-11e9-8dbd-bff5eb43a373.png)

It's not exactly the same appearance compare to the `gnuplot` generated one. I just wondering if this is acceptable. If yes, I am more than happy to make a PR that allows `criterion` plotting with `plotters` crate. Otherwise, would you mind provide some information about what is missing in the result.

FYI, the change I made for the prototype: https://github.com/38/criterion.rs/commit/1fae1a52e6958d01f5698bd829c060e6719601b1

bheisler:
Hey, thanks for getting in touch. I'm glad that Criterion.rs is working for you.

It doesn't matter too much if the `plotters` charts don't look exactly the same as the `gnuplot` ones. They should be readable and contain the same information. This looks fine to me.

I probably won't get around to implementing it myself any time soon, but a pull request would be welcome.

38:
Hi, 

I am decide to spend a little time on this task. So before I formal started, I probably need some input from you. 

I think I would like to first refactor code under `src/plot` to make the `criterion_plot` details hide behind a trait object and make `Html` relies on dynamic dispatch for plotting. 

[...]

PR #350, opened 2019-10-31
Plotters integration

38:
Hi, just as we discussed under issue #313 , this PR and this is very WIP in a very early stage. Any comments are welcome :)

- [x] Refactor `criterion::plot` code,  using dynamic dispatch for generating plots in `Html` code.
    - [ ] Extract the common data processing logic to the `PlotContext` and `PlotData` struct and reuse them for all backends
- [x] `Plotters` backend implemnetaion
    - [x] Distributions
    - [x] PDFs
    - [x] T-Test
    - [x] Regressions
    - [x] Line Comp and Violin 
- [x] Commandline argument for switching backends


bheisler:
Great! Thanks for helping out. I've skimmed the code changes and they look pretty reasonable so far.

I did notice that you're using a git dependency on `plotters` - that will need to be changed to a regular crate dependency before this is merged. Crates.io doesn't allow crates to have git dependencies.

38:
> Great! Thanks for helping out. I've skimmed the code changes and they look pretty reasonable so far.
> 
> I did notice that you're using a git dependency on `plotters` - that will need to be changed to a regular crate dependency before this is merged. Crates.io doesn't allow crates to have git dependencies.

Yep, there's a lot of things need to be clean up (But I am going to resolve them when it's done). 
One of the major reason for using latest version is I am also thinking about making change from Plotters side to make this PR more reasonable and maybe add essential features to Plotters as well.  Once I finished the PR ending up with new features added into Plotters, I think I can publish a new version of Plotters and make Criterion dependents on that version.

38:
Seems I just got the PR fulling functioning, but I think there's still few things needs to be cleanup. If you have any thoughts please leave comments. Thank you ! 

-----

As you mentioned before, you concerned about adding Plotter would introduce other dependencies that slow down the compilation. I believe it's fine to making plotters a required dependencies as Plotters won't introduce too much dependencies if only SVG backend is enabled.(The only required one is num-trait).

```
$ cargo tree -p plotters
plotters v0.2.11
├── num-traits v0.2.9
│   [build-dependencies]
│   └── autocfg v0.1.7
└── svg v0.6.0
```

svg is a pretty tiny crate, plus plotters itself doesn't takes too long time to compile, I believe the additional compile time should be less than 5s. And `num-traits` is a dependency of criterion already. 

-------

Just realized svg crate actually compiles pretty slow, so I just remove this dependency completely. Now I believe there's no significant slow down for introducing plotters deps.
image
Uploads the timing plot generated by cargo

Cargo.toml
```diff
@@ -32,6 +32,12 @@ rand_core = { version = "0.5", default-features = false }
 rand = "0.7"
 rayon = "1.1"
 
+[dependencies.plotters]
+# This should be changed to Plotters' next release
```
38: TODO: Change this to next published version

bheisler:
Hey, thanks for your patience. This all looks great! I'll probably do some minor tweaks after it's merged. Thanks for the (awesome!) pull request!

PR #359, opened 2019-11-27
Fix violin plot

### LINKED ISSUE #348: Malformed violin plot when comparing functions
Hello there,

After upgrading to criterion-0.3 I've found that comparison report doesn't work as expected anymore :( On ["fibonacci" comparison](https://bheisler.github.io/criterion.rs/book/user_guide/comparing_functions.html) I'm getting the following result:

image

And that's what I get when running gnuplot on generated `violin.gnuplot` manually:
image

That's totally weird and I'm a little bit lost. Might it be that I've got an "incomatible" gnuplot version or something? It's 5.3 on my machine (also tested with 5.2.7 with the same results).

Thanks!

Timmmm:
This fixes two issues:

* The y height of each plot was scaled according to the maximum x value.
* The x values were not scaled to match the unit used.

Fixes #348.

This is how it looks after this patch:

image

### DIFF
--- a/src/plot/summary.rs
+++ b/src/plot/summary.rs
@@ -159,8 +159,11 @@ pub fn violin(
             max = e;
         }
     }
-    let mut dummy = [1.0];
-    let unit = formatter.scale_values(max, &mut dummy);
+    let mut one = [1.0];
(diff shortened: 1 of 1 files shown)


bheisler:
Merged. Thanks for the fix!

PR #368, opened 2020-01-05
Add option to load data instead of sampling it

### LINKED ISSUE #357: Allow comparative analysis of two baselines
## Motivation

Criterion's elaborate statistical analysis can only be run on a *current, newly performed* benchmark against some previously established baseline measurement. It would also be useful to be able to do a comparison without having to do a new expensive benchmark run and instead utilize two saved measurements/baselines.

## Use cases

Two main use cases come to mind:

1. Compare one state of a program against multiple other previous runs such as against the integration branch, against master, and against the last release. But this may also apply to experimentation where several competing ideas should be compared against each other instead of the linear fashion that `baseline` suggests.

2. Allow the analysis to occur on a separate (virtual) machine than the performance measurement. For example, when running the performance capture in a separate machine to ensure the better control of the hardware and the runtime environment. But the output is persisted into the `target` directory which is typically wiped in CI, and the comparison analysis is not saved at all. A current workaround is to capture a the full analysis run but this restricts it to comparing the exact chosen baseline against that CI run.

197g:
Adds a new application argument:

`--load-baseline <name>`

This restores raw data from the implementation defined raw dumps and
subsequently treats it as if it had just been sampled. The dumps are
created for all runs early after the sampling and analysis process.
Note that this requires all benchmarks that are analyzed to have the
named baseline previously saved with --save-baseline (except the last
run which can be loaded under the name 'new').

The expose command line option requires a comparison baseline to be
specified as the option otherwise offers little advantage over the
already available csv exporter.

Open to discussion of the approach and alternatives. Also, the error 
behaviour could be adjusted, it does not feel correct yet. A mechanism to
actually import/export sample dumps can be discussed in future issues and 
PRs or implemented by external tools as filesystem operations.

Closes: #357

### DIFF
--- a/src/analysis/mod.rs
+++ b/src/analysis/mod.rs
@@ -82,14 +82,37 @@ pub(crate) fn common<M: Measurement, T>(
         return;
     }
 
-    let (iters, times) = routine.sample(
-        &criterion.measurement,
-        id,
-        config,
-        criterion,
-        report_context,
-        parameter,
-    );
+    let (iters, times);
+    if let Some(baseline) = &criterion.load_baseline {
+        let loaded = fs::load::<(Box<[f64]>, Box<[f64]>), _>(&format!(
+            "{}/{}/{}/sample.json",
+            criterion.output_directory,
+            id.as_directory_name(),
+            baseline
+        ));
+
+        match loaded {
+            Err(_) => panic!(format!(
+                "Baseline '{base}' must exist before it can be loaded; try --save-baseline {base}",
+                base = baseline,
+            )),
+            Ok(samples) => {
+                iters = samples.0;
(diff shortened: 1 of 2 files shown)


197g:
Example interaction:

```bash
# Prepare two baselines
git check master
cargo bench -- --save-baseline master
git checkout feature
cargo bench -- --save-baseline feature
# Do some feature work
cargo bench

# Create progress report of most recent benchmark against both feature branch and master
cargo bench -- --load-baseline new --baseline feature
cargo bench -- --load-baseline new --baseline master
```

bheisler:
This seems good to me. Thanks for the pull request!

Issue #355, opened 2019-11-25
Perform multiple Measurements at the same time


oberien:
Currently, I've only found ways to perform a single measurement at the same time. However, I'd like to measure both the cpu cycles (`rdtscp`) and time (`CLOCK_THREAD_CPUTIME_ID`) at the same time.
I could have two benchmarks, one with cycles and one with time, but I don't want to run the benchmarks twice as they already take in the range of 10 minutes.

I tried implementing a custom `Measurement`, which uses a tuple of `(cycles, time)` as `Value`. However, it appears that the measurement value will be converted to an `f64` for internal analysis and is pass to the `ValueFormatter` as an f64 as well.

I could think of the following two solutions:
* Let the formatter format the output by passing the original `Measurement::Value` types.
* Add a function `Criterion::with_measurements(&[&dyn Measurement])`, which allows performing multiple measurements during a single run.

bheisler:
Hey, thanks for trying Criterion.rs and thanks for the suggestion!

This was deliberately left out of the custom-measurement implementation. Measuring multiple things at once raised too many follow-on changes to other parts of the code. I'd need to change a bunch of types through the rest of the code, there would need to be separate sections in the reports for each measurement, CSV and JSON files would have to be modified, etc. etc. To be clear, I would still like to do it, but it will take a bunch of work so it probably won't happen any time soon - I haven't had much time to work on Criterion.rs lately.

rrnewton:
These days these kinds of refactorings may be much easier with agents helping. I'll give it a brief try and see what the situation is.

I just bumped against this because I was appalled my claude agent was building duplicative benchmarks instead of taking multiple measurements from one (time, bytes allocated).

PR #369, opened 2020-01-08
Add Criterion::measurement function

oberien:
I wanted to write a benchmark function generic over the measurement such that I can easily benchmark once using cycles and once using time (see #355). However, there was no way to generically perform the measurements, which is why I added this function. The clone might not necessarily be required, but otherwise lifetime foo would become complicated, and usually it's a unit-struct anyway.

See the example in the doc-comment for an example use-case.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1336,6 +1336,7 @@ impl<M: Measurement + Clone> Criterion<M> {
     ///     bench("cycles", &mut criterion_cycles);
     /// }
     /// ```
+    #[allow(clippy::needless_doctest_main)]
     pub fn measurement(&self) -> M {
         self.measurement.clone()
     }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1287,7 +1287,7 @@ impl<M: Measurement + Clone> Criterion<M> {
     ///
     /// fn bench(c: &mut Criterion<impl Measurement + Clone + 'static>) {
     ///     let measurement = c.measurement();
-    ///     c.bench_function("iter", move |b| {
+    ///     c.bench_function("measurement", move |b| {
     ///         b.iter_custom(|iters| {
     ///             let input = black_box(1337);
     ///             let start = measurement.start();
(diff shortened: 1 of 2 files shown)


bheisler:
Huh, that's an interesting idea. I hadn't thought of that! One problem I can foresee is that you now have defined multiple benchmarks with the same name, and those will conflict.

oberien:
The benchmark `"iter"` is already defined twice in different doc-comments, if that's what you mean (also `"function_name"` is used multiple times):

https://github.com/bheisler/criterion.rs/blob/72a27c4742171999ef609ad4bb7b1cf15f0e4768/src/lib.rs#L312-L316

https://github.com/bheisler/criterion.rs/blob/72a27c4742171999ef609ad4bb7b1cf15f0e4768/src/lib.rs#L355-L365

I've changed it anyways.

If instead you mean that calling the `bench` function multiple times in my example results in weird behaviour, the main reason to make `Criterion` generic for me is to have two different benchmarks (as in benchmark binaries) with different measurements, in which case it's also fine to use the same name. Otherwise one can pass the name as argument to that function.

bheisler:
Hey, sorry I was unclear. What I meant was this:
`c.bench_function("measurement", move |b| {`
If you call this function with two or more different measurements, you've created two or more different benchmarks with the ID "measurement". The benchmark ID must be unique between _all_ of the different benchmarks in all of the benchmark binaries in your project because it's used as part of the path to the directory that stores the results. If you reuse IDs, the benchmarks will overwrite each other's measurements and have other strange behavior. As far as I can tell, there is no reliable way for Criterion.rs to detect this to warn the user.

On the other hand, I suppose that if you're using `iter_custom` in the first place you're probably a more advanced user, so maybe it's OK. The documentation should warn about that and provide a correct example though.

oberien:
I added a more complex example showcasing measurement using both `WallTime` and (stubbed) `Cycles`, which pass different names to the bench function. I hope this makes it clearer that different names should be used.

Maybe it would make sense to add a note to `Criterion::bench_function` and `BenchmarkGroup::bench_function` stating that names must be unique over all benchmarks in the project.

lemmih:
This looks useful to me. Are you up for resolving the conflicts?

lemmih:
Closing due to inactivity.

oberien:
Sorry for the late response. I've rebased my fork to the latest master.

lemmih:
I can't reopen after you're rebased your branch. You'll have to create a new PR.

PR #372, opened 2020-01-29
Fix cargo bench flag

nkconnor:


nkconnor:
I didn't read enough :)

PR #373, opened 2020-01-29
Set `default-features = false` for  `num-traits`

Pratyush:
Hi!

[We're](https://github.com/scipr-lab/zexe/pull/76#issuecomment-579924945) running into https://github.com/rust-lang/cargo/issues/4866 because `criterion` uses `num-traits` with the default features, which turns off `no-std` mode. This PR turns off the default features and (hopefully) fixes that. Thanks!


bheisler:
Hey, thanks for your patience. This seems good to me. Thanks for the pull request!

PR #378, opened 2020-02-15
Fix homebrew install in Gnuplot builds

palfrey:
Not quite sure when the `brew update` was needed, but it doesn't seem to be required now. More importantly, the core problem was that `python@2` was already linked, and we need to unlink it to make the homebrew install work properly as we're installing Python 3 as part of the dependencies.

### DIFF
--- a/ci/install.sh
+++ b/ci/install.sh
@@ -10,7 +10,6 @@ if [ "$DOCS" = "yes" ]; then
 fi
 
 if [ "$TRAVIS_OS_NAME" = "osx" ] && [ "$GNUPLOT" = "yes" ]; then
-    brew update
     brew install gnuplot
 fi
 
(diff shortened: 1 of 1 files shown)


bheisler:
Hey, thanks for the pull request. I don't know much about MacOS so this is really helpful.

palfrey:
FYI the build issue now turning up (https://travis-ci.org/bheisler/criterion.rs/jobs/656710537) is an upstream issue https://github.com/Michael-F-Bryan/mdbook-linkcheck/issues/27

palfrey:
> FYI the build issue now turning up (https://travis-ci.org/bheisler/criterion.rs/jobs/656710537) is an upstream issue [Michael-F-Bryan/mdbook-linkcheck#27](https://github.com/Michael-F-Bryan/mdbook-linkcheck/issues/27)

This should work again now https://github.com/TeXitoi/structopt/issues/354 is fixed and a new version of structopt is out.

PR #384, opened 2020-02-29
Avoiding dropping Vec inside the measurement in iter_bached

gereeter:
The deallocation is probably not a huge overhead, but it is unnecessary. `drain` is stable since Rust 1.6.0.


lemmih:
Hi @gereeter. I'm a new maintainer for criterion and your change looks good. The source code has been moved around, unfortunately, and your PR needs to be updated before it can be merged.

I'll totally understand if you're not keen on putting in more work after all this time. In any case, I'll close this PR after one week.

lemmih:
Closing due to inactivity.

Issue #376, opened 2020-02-02
Support for subtractive timing model?


bluenote10:
I've always been wondering if it wouldn't make more sense to benchmark very fast operations with a subtractive timing model against a task specific reference noop. The logic of the timing loop would be:
- measure `iters` iterations of executing a `noop` function.
- measure `iters` iterations of executing the `routine` function.
- return difference of `t_routine - t_noop`.

In terms of the timing model this would lead to a cancellation of the overhead terms:

```
elapsed = + (Instant::now + iters * (routine + mem::drop(O) + Range::next))
          - (Instant::now + iters * (noop + mem::drop(O) + Range::next))
        = iters * (routine - noop)
```

Here is an experimental implementation:

```rust
use criterion::{black_box, criterion_group, criterion_main, Criterion};
use std::time::{Duration, Instant};

fn bench_function_with_noop<O, N, R>(iters: u64, mut noop: N, mut routine: R) -> Duration
where
    N: FnMut() -> O,
    R: FnMut() -> O,
{
    let start = Instant::now();
    for _i in 0..iters {
        black_box(noop());
    }
    let t_noop = start.elapsed();

    let start = Instant::now();
    for _i in 0..iters {
        black_box(routine());
    }
    let t_routine = start.elapsed();

    // Durations can only be positive. We have to live with a bias 
    // of truncating for now...
    if let Some(diff) = t_routine.checked_sub(t_noop) {
        diff
    } else {
        std::time::Duration::from_nanos(0)
    }

}

// Example usage:

#[inline]
fn noop<T>(x: T) -> T {
    x
}

fn custom(c: &mut Criterion) {
    c.bench_function("test", |b| b.iter_custom(|iters| {
        bench_function_with_noop(
            iters,
            || noop(black_box(1.0_f64)),
            || black_box(1.0_f64).abs(),
        )
    }));
}

criterion_group!(benches, custom);
criterion_main!(benches);
```

Does such an approach make sense? If so, could we properly support it by allowing for negative durations in the interface, or even offering versions of `iter` and `iter_batched` with noops?

-----

Overall I have the impression that the subtractive timing model gives quite reasonable results. For instance, the following measures integer additions both with the standard model and the subtractive model:

```rust
c.bench_function("add_one_op (standard)", |b| b.iter(
    || black_box(1) + black_box(1)
));
c.bench_function("add_ten_ops (standard)", |b| b.iter(
    || black_box(1) +
        black_box(1) +
        black_box(1) +
[...]

PR #385, opened 2020-02-29
Use a subtractive timing model in iter_batched{,_ref}

gereeter:
See #376. This increases variance by doubling the number of measurements, but removes the bias due to iteration and measurement (per-batch) overhead. `iter_batched` already is using more measurements to reduce bias, so this seems reasonable.

This is based on #384 because I thought of that first and it also is about removing bias from those functions.

I'm a bit confused about what the correct thing to put in the `Timing model` documentation section is. I adjusted it to describe the expectation of the recorded `value - overhead`, which seemed like the most useful interpretation that fits the existing documentation, but `O::drop` was in the documentation for `iter_batched`, yet the drop of the output is very carefully *not* part of what is measured. Also, only `iter` seems to mention the `Range::next` overhead.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -282,6 +282,7 @@ pub struct Bencher<'a, M: Measurement = WallTime> {
     iterated: bool,         // have we iterated this benchmark?
     iters: u64,             // Number of times to iterate this benchmark
     value: M::Value,        // The measured value
+    overhead: M::Value,     // The measured overhead, to be subtracted from the value
     measurement: &'a M,     // Reference to the measurement object
     elapsed_time: Duration, // How much time did it take to perform the iteration? Used for the warmup period.
 }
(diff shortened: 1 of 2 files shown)


lemmih:
Interesting PR. Would you be up for resolving the conflicts?

bheisler:
I think I'd like to see more discussion justifying this change to the timing model before I'd want to merge something like this. Timing things precisely is harder than it seems; there is a real possibility of overestimating the time taken by overhead and incorrectly reporting that the benchmarked code is faster than it actually is. Optimizing compilers make this even harder; you can't just naively iterate the overhead code without the benchmark code to measure that separately, because the compiler may make different optimization decisions for the overhead-without-benchmark than it did for the overhead-with-benchmark.

lemmih:
Closing due to inactivity.

PR #387, opened 2020-03-05
Add reporter format with bencher support

pksunkara:
This allows us an option print bencher type format on CLI

```
$ cargo bench -- -R bencher
   Compiling criterion v0.3.1 (/Users/pksunkara/Coding/clap-rs/criterion.rs)
   Compiling clap v3.0.0-beta.1 (/Users/pksunkara/Coding/clap-rs/clap)
    Finished bench [optimized] target(s) in 1m 15s
     Running target/release/deps/01_default-439bccc0e739a77b
Gnuplot not found, using plotters backend
test build_empty_app ... bench:        135 ns/iter (+/- 0)                                 
test parse_empty_app ... bench:      1,940 ns/iter (+/- 10)  
```

### DIFF
--- a/src/benchmark.rs
+++ b/src/benchmark.rs
@@ -306,6 +306,7 @@ impl<M: Measurement> BenchmarkDefinition<M> for Benchmark<M> {
             output_directory: c.output_directory.clone(),
             plot_config: self.config.plot_config.clone(),
             test_mode: c.test_mode,
+            reporter: c.reporter.clone(),
         };
 
         let config = self.config.to_complete(&c.config);
@@ -455,6 +456,7 @@ where
             output_directory: c.output_directory.clone(),
             plot_config: self.config.plot_config.clone(),
             test_mode: c.test_mode,
+            reporter: c.reporter.clone(),
         };
 
         let config = self.config.to_complete(&c.config);
--- a/src/benchmark_group.rs
+++ b/src/benchmark_group.rs
@@ -270,6 +270,7 @@ impl<'a, M: Measurement> BenchmarkGroup<'a, M> {
             output_directory: self.criterion.output_directory.clone(),
             plot_config: self.partial_config.plot_config.clone(),
             test_mode: self.criterion.test_mode,
+            reporter: self.criterion.reporter.clone(),
         };
 
         let mut id = InternalBenchmarkId::new(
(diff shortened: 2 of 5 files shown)


pksunkara:
@bheisler I don't why only that 1 job is failing.

pksunkara:
Looks like it is failing on master also.

bheisler:

I'm not opposed to it, but I would have preferred some discussion _before_ you implemented it. Why do we need to emulate bencher's output? It provides strictly less information than Criterion.rs' existing output, so when would someone want to enable this?

pksunkara:
So, I know it provides less information. But adding this reporter makes criterion compatible with benchmarking tools that support bencher. We wanted to use https://github.com/rhysd/github-action-benchmark in https://github.com/clap-rs/clap. And this is an easier but still correct way of doing it rather than trying to add criterion specific support to that tool.

pksunkara:
@bheisler Can I help with something to prioritise this? I would very much like to see this done.

bheisler:
Thanks for your patience; I've been busy for a while.

Thanks for the suggestion! I've decided to implement this change myself in a slightly different way. Could you check out the master branch and see if it works for you? The output format isn't _quite_ the same as `bencher`, but it should be close enough that supporting it in the github action will be easy. The command line argument is `--output-format bencher`.

I've tried to set it up so that the `bencher` format should print the measurements in a consistent format with a consistent unit, but there are external crates that provide custom measurements like cycles per iteration. Those crates may require some extra work on the github action to support.

pksunkara:
The code looks good. I am currently testing the output format.

pksunkara:
You are missing ` bench` on [this](https://github.com/bheisler/criterion.rs/commit/77ebdf3b689e76a5db9260e2bc7f7ba3465a1c35#diff-5e58a0d0718c8356bc1e952721fb72e4R754) line. Check the extraction regex at, https://github.com/rhysd/github-action-benchmark/blob/master/src/extract.ts#L181

pksunkara:
#### Your master

```
test build_empty ...:         139 ns/iter (+/- 1)

test parse_empty ...:        2069 ns/iter (+/- 144)

```

#### My branch

```
test build_empty ... bench:        139 ns/iter (+/- 0)
test parse_empty ... bench:      2,062 ns/iter (+/- 46)
```

bheisler:
Whoops, my bad. Should be fixed now.

pksunkara:
@bheisler Can confirm the master branch works. https://github.com/clap-rs/clap/actions/runs/87790559. Hoping you can release something soon.

bheisler:
Awesome, thanks.

pksunkara:
Anything I can help with which is blocking the next release?

bheisler:
Just published 0.3.2.

PR #389, opened 2020-03-13
Update documentation on the MSRV

LingMan:
Currently Rust 1.33 or later is required to build Criterion.rs. Update documentation suggesting older versions work.


bheisler:
Thanks for the pull request!

PR #390, opened 2020-03-15
Disable as_deref lint as it doesn't work until Rust 1.40.0

palfrey:
And the current MSRV here is 1.33


bheisler:
Thanks for the pull request, but I've made this change myself already.

PR #391, opened 2020-03-17
Avoid panic if environment contains non-UTF8 var

Andlon:
I wanted to use google's CPU profiler with `criterion`, but I often find it very practical to just add it using `LD_PRELOAD` instead of linking it into the library. This normally works rather well. However, it turns out that `cpuprofiler` actually clobbers the environment variable `CPUPROFILE` instead of just reading it (see here: https://github.com/gperftools/gperftools/blob/f47a52ce85c3d8d559aaae7b7a426c359fbca225/src/base/sysinfo.cc#L209), resulting in a non-utf8 variable. This in turn makes `std::env::vars` panic when iterating, and consequently `criterion` panics as well.

`vars` is used two places in `criterion`: one to determine if we should enable debugging, and one for determining the cargo target directory. Fixing the first one is easy. A proper fix for the second instance would require us to change all paths used in the `Criterion` struct from `String` to `PathBuf`. I have not done this in this PR: I could do so, but I thought I'd ask your opinion first. Instead, as a hotfix, I've made sure that we don't panic if the environment contains an unrelated environment variable that is not UTF-8 compatible.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -100,7 +100,7 @@ pub use crate::benchmark::{Benchmark, BenchmarkDefinition, ParameterizedBenchmar
 pub use crate::benchmark_group::{BenchmarkGroup, BenchmarkId};
 
 lazy_static! {
-    static ref DEBUG_ENABLED: bool = { std::env::vars().any(|(key, _)| key == "CRITERION_DEBUG") };
+    static ref DEBUG_ENABLED: bool = { std::env::var("CRITERION_DEBUG").is_ok() };
     static ref GNUPLOT_VERSION: Result<Version, VersionError> = { criterion_plot::version() };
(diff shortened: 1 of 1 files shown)


bheisler:
Hey, thanks for the pull request and thanks for your patience.

I've opted to implement this myself instead, rather than make you update it. I've gone a bit further than you have; I've made it handle non-UTF8 characters in the two environment variables that Criterion.rs looks at, and changed the output directory to a PathBuf. To avoid a breaking change, baseline names must still be valid UTF8, and are still represented as a String.

Could you try using the latest master version of Criterion.rs and let me know if it worked? Thanks in advance!

Andlon:
Thanks @bheisler, that's great! I just tried out the latest release, and it seems to work flawlessly. Awesome :-)

PR #393, opened 2020-03-20
Update plots_and_graphs.md

humb1t:
Different names for the same folder may be misleading. Renamed `$BENCHMARK` into `$BENCHMARK_NAME`.


bheisler:
Thanks for the pull request!

PR #394, opened 2020-03-22
Lint fixes

g2p:
The first commit fixes a warning that prevented building on current nightlies.
The rest add fixes for some other clippy and edition lints.


g2p:
The first fix, for the warning, requires [Rust 1.40](https://caniuse.rs/features/option_deref).
This is okay by the stated MSRV policy (last three stable releases), so I will update the Travis script if maintainers agree.
Alternatively, the warning that would have to be whitelisted is called clippy::option_as_ref_deref.

bheisler:
Thanks for the pull request, but I've fixed the lints myself.

PR #399, opened 2020-04-01
fix parameter and imports

Satoshi-Kusumoto:


bheisler:
Thanks for the pull request! Merged.

PR #401, opened 2020-04-10
Allow dynamically sized parameter as input

197g:
All functions with user supplied paramter only deal with it behind a reference. This allows passing a slice or a `&str` as a parameter.


bheisler:
Huh. Yeah, that makes sense. Surprised I haven't thought of this before now. Thanks for the pull request!

PR #416, opened 2020-05-19
Replaced tempdir crate with tempfile

jbreitbart:
RUSTSEC-2018-0017: tempdir: `tempdir` crate has been deprecated; use `tempfile`
instead

See https://github.com/jbreitbart/criterion-perf-events/issues/2


bheisler:
Not particularly concerned about using a deprecated crate in my unit tests, but fair enough I suppose. Thanks for the pull request.

PR #425, opened 2020-06-29
Add default_mode_bench and default_mode_test functions

exists-forall:
For some use cases, it is desirable to be able to use Criterion.rs in a standalone executable invoked directly from the command line, bypassing `cargo bench`.  For example, I am using Criterion.rs to benchmark the code generated by an optimizing compiler, and I want to run these benchmarks using a debug build of my compiler to minimize build times.  `cargo bench` would force me to compile my compiler in `--release` mode, so I am using Criterion.rs from a standalone Rust executable instead, not managed by `cargo bench`.

For this use case, it is slightly inconvenient that one needs to pass `--bench` to the benchmark executable in order to run Criterion.rs in benchmarking mode (as opposed to test mode).  This pull request adds a configuration function to the `Criterion` struct that allows the user to specify `--bench` as the default execution mode for Criterion.rs, even if `--bench` is not explicitly passed on the command line.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -665,6 +665,12 @@ pub enum PlottingBackend {
     Plotters,
 }
 
+#[derive(Debug, Clone, Copy)]
+enum DefaultCliMode {
+    Benchmark,
+    Test,
+}
+
 #[derive(Debug, Clone)]
 /// Enum representing the execution mode.
 pub(crate) enum Mode {
@@ -715,6 +721,7 @@ pub struct Criterion<M: Measurement = WallTime> {
     measurement: M,
     profiler: Box<RefCell<dyn Profiler>>,
     connection: Option<MutexGuard<'static, Connection>>,
+    default_cli_mode: DefaultCliMode,
     mode: Mode,
 }
 
@@ -775,6 +782,7 @@ impl Default for Criterion {
             connection: CARGO_CRITERION_CONNECTION
                 .as_ref()
                 .map(|mtx| mtx.lock().unwrap()),
+            default_cli_mode: DefaultCliMode::Test,
             mode: Mode::Benchmark,
         }
     }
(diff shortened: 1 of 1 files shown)


bheisler:
Hey, thanks for the suggestion!

I just want to make sure I understand here - you're generating and benchmarking optimized code at runtime. You want to avoid optimizing the code generator itself to reduce iteration time, so you're running the un-optimized binaries as benchmarks manually because `cargo bench` doesn't provide an option for benchmarking un-optimized binaries. You're running them directly on the command line, and you'd rather not have to pass `--bench` when you do.

Now that you mention it, it is sort of weirdly asymmetric how `cargo test --release` exists but `cargo bench --debug` doesn't.

I'm skeptical of expanding the API surface and adding functionality that will have to be supported for a long time, and for such a small improvement on such a niche use case. It doesn't seem like a good tradeoff. I think there are better ways to solve this problem.

I think one way would be to add a `--debug` option to [cargo-criterion](https://github.com/bheisler/cargo-criterion). Since cargo-criterion invokes the compiler as well as running the benchmarks, it's trivial for it to build the benchmarks in debug mode. I also feel like a command-line option is the right user experience - that way you don't have to change your source code and don't have to complicate Criterion.rs' API. I'm planning to do the first alpha release of `cargo-criterion` today, in fact, and I've just added a `--debug` flag to it.

Another option would be to edit your Cargo.toml file to disable optimization of the `bench` profile, like so:

```toml
[profile.bench]
opt-level = 0
```

That will disable optimizations when running `cargo bench`, and then you can just run that instead of manually launching the benchmark commands. See [the Cargo documentation](https://doc.rust-lang.org/cargo/reference/profiles.html) for more information on that. If you have a new version of Cargo you can even configure it to compile your dependencies with optimizations but leave your own crate un-optimized, for a nice combination of recompile time and execution speed ([see here](https://doc.rust-lang.org/cargo/reference/profiles.html#overrides)).

lemmih:
Hi @exists-forall. Sounds like @bheisler suggested several workable solutions so I'm closing this PR. If criterion/cargo-criterion still doesn't fit your use case, please open an issue.

PR #439, opened 2020-09-05
Migrate to plotters 0.3

38:
Hi there,

I am just published Plotters 0.3 and in the future I would like to focus on maintaining the Plotters 0.3. So I would like make criterion depends on a 0.3 version of Plotters which will be actively maintained. 

Plotters 0.3 has some breaking changes compare to Plotters 0.2, this PR just made those changes. I am just tried Criterion with Plotters 0.3, it works well for me.

For the compile time, the new Plotters v0.3 seems fine (Plotters 0.3 cost ~3s to compile). 

Thanks!

- Hao

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,9 +33,9 @@ rayon = "1.3"
 regex = { version = "1.3", default-features = false, features = ["std"] }
 
 [dependencies.plotters]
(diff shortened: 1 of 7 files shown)


bheisler:
Thanks for the pull request!

PR #456, opened 2021-01-03
Fix report directory in Cargo workspace

### LINKED ISSUE #192: Reports are put into incorrect directory when used in a workspace

rossmacarthur:
Resolves #192

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -73,11 +73,13 @@ mod stats;
 use std::cell::RefCell;
 use std::collections::HashSet;
 use std::default::Default;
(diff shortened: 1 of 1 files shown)


bheisler:
Thanks for the pull request!

PR #474, opened 2021-05-04
Corrects `Criterion.toml` in the book

davidkern:
Corrects a small, though important, typo in the book.  It took a while to figure out why my plot colors were not changing. 😆


bheisler:
Hello! Thanks for the pull request, and good catch!

PR #477, opened 2021-05-10
Bug in variance calculation for `Sample`

yoniko:
The `var` function in `Sample` divides by `n-1` instead of by `n`.
As a result we get the wrong variance.

PR implements the following:
1. Bugfix
2. Since variance's calculation doesn't divide by len-1 anymore we can have a Sample with a single value
3. Added testing for the variance code


yoniko:
See issue #478

Eh2406:
Why do you think the population variance (`/n`) is more appropriate then the sample variance (`/(n-1)`) in this case?

yoniko:
@Eh2406 - you are correct, I guess my statistics is rusty and I failed to realize the difference.
Closing the PR.

Issue #467, opened 2021-02-15
plot/plotters_backend: infinite loop when all samples are the same


elagergren-spideroak:
This snippet below will cause an infinite loop when generating `pdf.svg`. Specifically, `plotters_backend::pdf` calls `Sample::new(&xs)`, which results in `xs_.min() == xs_.max()`. The variable `xs_` is passed to `set_secondary_coord`, which generates a `Cartesian2d`. On line 232, `MeshStyle.draw` is called, which then calls `target.draw_mesh`, which tries to create `x_label` and `y_label` by calling `self.draw_mesh_lines`. That, in turn, calls `self.drawing_area.draw_mesh`, which calls `self.coord.draw_mesh`. Here, `self.coord` is `Cartesian2d` (as generated by `set_secondary_coord`). Calling its `draw_mesh` method tries to create `xkp` and `ykp` by calling `self.logic_x.key_points` and `self.logic_y.key_points`, respectively. Those methods are generated by the `gen_key_points_cmp` macro, which never breaks the `'outer` label. The label is never broken because:

1. `scale` is zero because `range.1 == range.0`: `10 ** floor(log10(x, x)) == 0`
2. `next_left` is NaN because `scale/nxt == 0`: `rem_euclid(a, 0) = a-ceil(a/0)*0 = a-ceil(+inf)*0 = a-NaN = NaN`
3. Same with `next_right`.
4. And NaN is sticky, so `npoints` is NaN.
5. `NaN > max_points` is false.

Admittedly, it's contrived: it requires a custom benchmark that always returns the same number. But, that could be the result of a bug, which is how I ran into it. And it's not exactly straightforward to debug. And while I haven't looked into it further, ISTM that it's even reproducible if N isn't a constant.

I'm also not sure if it's a bug in Criterion or in `plotters`. I'm not sure if a zero coordinate is valid or not. 

```
use criterion::{
    criterion_group, criterion_main,
    measurement::{Measurement, ValueFormatter},
    Criterion, Throughput,
};
use std::time::Duration;

fn repro(c: &mut Criterion<MyMeasurement>) {
    c.bench_function("repro", |b| {
        b.iter_custom(|_| {
            return 42;
        })
    });
}

struct MyMeasurement;

impl Measurement for MyMeasurement {
    type Intermediate = u32;
    type Value = u32;

    fn start(&self) -> Self::Intermediate {
        return 0;
    }
    fn end(&self, i: Self::Intermediate) -> Self::Value {
[...]

PR #481, opened 2021-05-22
Upgrade Plotters to v0.3.1

38:
Upgrade Plotters to v0.3.1, which including the fix for infinite loop bug with the zero-length axis is passed in.  This should address issue #467.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,7 +37,7 @@ tokio = { version = "1.0", default-features = false, features = ["rt"], optional
 async-std = { version = "1.9", optional = true }
 
 [dependencies.plotters]
-version = "^0.3.0"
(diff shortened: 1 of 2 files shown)


mb:
Thanks writing the fix. I've tested it and it works for my case. I would really appreciate if this could get merged, since then I don't have to depend on my local copy of criterion.

The NaN data points I described in #467 weren't related to the issue. They were caused by returning a duration of 0 when 0 iterations were requested. Should I open another issue for that?

lemmih:
Ping @bheisler

lemmih (approved):

lemmih:
Looks good to me. No breaking changes so this should be in the next minor release.

PR #488, opened 2021-07-25
Use GitHub Actions to test with MSRV + stable + beta + nightly.

lemmih:


lemmih:
Sigh, `walkdir` accidentally bumped their MSRV to 1.40 without noticing it.

PR #492, opened 2021-07-26
Partly disable appveyor until performance can be improved.

lemmih:
The Appveyor build requires 20-30 minutes to complete which limits feedback on PRs. With caching, it should take no more than 5 minutes to build and test all of `criterion`. Unfortunately, I'm not an expert at configuring Appveyor so we'll rely on GitHub Actions until I figure out the caching issue.


lemmih:
Hm, renaming `appveyor.yml` didn't do the trick.

lemmih:
Hm, it might be the `cargo bench` command that takes most of the build time.

lemmih:
Building + Testing: ~5 minutes.
Benchmarking: 10 minutes.

There are two build configurations so the total time is roughly 30 minutes.

PR #541, opened 2021-12-27
Add a `cargo_criterion_support` feature flag

str4d:
This enables `cargo-criterion` support to be compiled out if it is not required. A compile-time error is added to require that at least one of [`cargo_bench_support`, `cargo_criterion_support`] must be enabled.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -27,6 +27,11 @@
     )
 )]
 
+#[cfg(not(any(feature = "cargo_bench_support", feature = "cargo_criterion_support")))]
+compile_error!(
+    "At least one of [`cargo_bench_support`, `cargo_criterion_support`] must be enabled"
+);
+
 #[cfg(test)]
 extern crate approx;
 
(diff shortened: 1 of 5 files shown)


src/lib.rs
```diff
@@ -123,6 +131,9 @@ lazy_static! {
             PlottingBackend::None
         }
     };
+}
+#[cfg(feature = "cargo_criterion_support")]
+lazy_static! {
     static ref CARGO_CRITERION_CONNECTION: Option<Mutex<Connection>> = {
         match std::env::var("CARGO_CRITERION_PORT") {
```
str4d, line 138: Currently, when `cargo_criterion_support` is not enabled, `criterion` behaves as if `cargo-criterion` does not exist. It might be useful to check somewhere whether the `CARGO_CRITERION_PORT` env variable is set, and print a warning or error if `cargo_criterion_support` is not enabled. This wouldn't be the right place; it would need to go in the benchmark's `main` function (similar to the previous warnings when `cargo_bench_support` was not enabled).

lemmih:
I'm not sure what we should do about `cargo_bench_support`. The benefits of using `criterion` without `cargo bench` support seem very minor to me. Unless there is a compelling use-case, I think we should keep `cargo_bench_support` as a dummy feature that does nothing.

PR #542, opened 2022-01-02
Fix compile error in iai example

wduminy:


lemmih:
Test failure isn't related to this PR. I will fix it.

PR #543, opened 2022-01-02
Upgrade to clap v3

benesch:
@Lemmih I know y'all sometimes discourage dependency bump PRs, but we'll be running a forked version with this patch @MaterializeInc. Figured I'd toss it up in case it saves you some work.

----

Clap v3 is a new major version of clap with several deprecations and
breaking changes, all of which are addressed within this commit.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -36,7 +36,6 @@ extern crate approx;
 #[cfg(test)]
 extern crate quickcheck;
 
-use clap::value_t;
 use regex::Regex;
 
 #[macro_use]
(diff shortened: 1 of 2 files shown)


lemmih:
Hi @benesch. I'd love to review the PR for the next major release. Mind targeting the `version-0.4` branch instead of `master`?

benesch:
Awesome, done! 

Out of curiosity, do you have a timeline for the release of v0.4?

lemmih:
> Awesome, done!
> 
> Out of curiosity, do you have a timeline for the release of v0.4?

Q1 of 2022.

lemmih:
Hmm, it seems like some of the new dependencies have an MSRV greater than 1.46. We'll have to consider bumping it for criterion.

benesch:
Ah yeah, looks like os_str_bytes requires Rust 1.52 or greater: https://github.com/dylni/os_str_bytes#rust-version-support

benesch:
Seems like you'd be well within your rights to bump to v1.52 for the next major release of Criterion.rs, per the README!

> Criterion.rs supports the last three stable minor releases of Rust. At time of writing, this means Rust 1.50 or later. Older versions may work, but are not guaranteed.
>
> Currently, the oldest version of Rust believed to work is 1.46. Future versions of Criterion.rs may break support for such old versions, and this will not be considered a breaking change. If you require Criterion.rs to work on old versions of Rust, you will need to stick to a specific patch version of Criterion.rs.

src/lib.rs
```diff
@@ -1009,56 +1008,54 @@ https://bheisler.github.io/criterion.rs/book/faq.html
...
         if matches.is_present("confidence-level") {
-            let num_confidence_level = value_t!(matches.value_of("confidence-level"), f64)
-                .unwrap_or_else(|e| {
-                    println!("{}", e);
-                    std::process::exit(1)
-                });
+            let num_confidence_level = matches.value_of_t("confidence-level").unwrap_or_else(|e| {
+                println!("{}", e);
+                std::process::exit(1)
+            });
```
coolreader18: I think for these you can use `matches.value_of_t_or_exit()`, right?

benesch: Oh yeah, looks like it. @Lemmih I'll leave that to you so I don't accidentally blow away whatever state you might have, unless you want me to make the change.

lemmih: You go ahead.

mati865: ping @benesch

benesch: Ok, done.

vthib:
We'd like to get this change on the next major version of criterion as well, it's the only dependency we have still depending on clap v2

lemmih (approved):

Issue #514, opened 2021-09-24
[feature request] Way to run setup only when benchmark is actually run.


rukai:
Currently I have many benchmarks that look like:

```rust
fn foo(c: &mut Criterion) {
    let setup = do_expensive_setup();

    c.bench_function("foo", move |b| {
        b.iter(|| {
            do_something_with(setup);
        })
    });
}
```

When I run a specific test via `cargo bench foo` then every do_expensive_setup is run regardless of if it is used by foo.

So to solve this issue it would be great if there was some way to run setup code only when the actual benchmark is run.
Maybe something like a `Criterion::bench_function_with_one_time_init`?
https://docs.rs/criterion/0.3.5/criterion/struct.Criterion.html#method.bench_with_input isnt suitable because it still needs the setup to always be run, I need to be to provide a closure so no work is done unless the bench is run.

Edit: Realized that the destructors are being run before moving onto the next bench mark so I removed the complaints about side effects.

rukai:
I'm going to start working on an implementation of this.

rukai:
I no longer believe this is a sensible thing for criterion to support.
criterion should focus on microbenchmarks, but we were trying to use it for high level database benchmarks.
Instead we have started developing our own database benchmarking framework: https://github.com/shotover/shotover-proxy/tree/main/windsock
As such I will close this issue.

PR #556, opened 2022-02-09
bench_with_input now takes an FnOnce instead of raw input

rukai:
closes https://github.com/bheisler/criterion.rs/issues/514

I chose to alter bench_with_input to add this functionality because it can still be used for the original use case by just adding `||` to make the value into a closure.
Having one multipurpose function instead of 2 keeps the API simpler.
It also encourages users to generate their input lazily when possible.
But I dont feel strongly about this and im happy to go for any other design that solves #514 that you might suggest.

Making the input provided to the bench function always mutable is needed so that we can use things like a connection type in the input.
For an example of this you can refer to https://github.com/shotover/shotover-proxy/pull/514/files

### DIFF
--- a/benches/benchmarks/compare_functions.rs
+++ b/benches/benchmarks/compare_functions.rs
@@ -27,18 +27,22 @@ fn fibonacci_fast(n: u64) -> u64 {
 fn compare_fibonaccis(c: &mut Criterion) {
     let mut group = c.benchmark_group("Fibonacci");
 
-    group.bench_with_input("Recursive", &20, |b, i| b.iter(|| fibonacci_slow(*i)));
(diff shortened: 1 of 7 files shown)


rukai:
ci failure is not my problem.

You cant fail ci on clippy lints unless its only running against a pinned rustc version.

rukai:
I no longer believe this is a sensible thing for criterion to support.
criterion should focus on microbenchmarks, but we were trying to use it for high level database benchmarks.
Instead we have started developing our own database benchmarking framework: https://github.com/shotover/shotover-proxy/tree/main/windsock
As such I will close this PR.

Issue #435, opened 2020-08-28
Text is cut off if the terminal width is not large enough


jyn514:
See for example the following output:
```
Benchmarking set allocating: Collecting 100 samples in estimated 5.0539 s (404k itera                                                                                     set allocating          time:   [13.266 us 13.757 us 14.321 us]
                        change: [+25.539% +31.449% +38.019%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 5 outliers among 100 measurements (5.00%)
  4 (4.00%) high mild
  1 (1.00%) high severe

Benchmarking set without allocating: Collecting 100 samples in estimated 5.0191 s (75                                                                                     set without allocating  time:   [10.939 us 11.464 us 12.067 us]
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high mild
```
The text after 80 columns is cut off completely, instead of being printed normally. This is not at all desired behavior - criterion should print it normally and let the terminal handle word wrapping. In particular this breaks copy pasting since the text is just not there.

lemmih:
Hi @jyn514. I'm not able to reproduce this behavior. For me, the output is wrapped by the terminal. Looking at the code, it appears to be ordinary `println!` and `eprintln!`.

Are you still seeing this behavior with criterion-0.3.5? If so, which terminal are you using?

jyn514:
Yes, I am still seeing this behavior with 0.3.5. I've used several different terminals and run into the same problem - currently I'm using `kitty`, but I think I had the same problem with Konsole.

Maybe this is some interaction between the bench framework and criterion? I'm running this with `cargo bench`.

```
Benchmarking set allocating: Collecting 100 samples in estimated 5.0050 s (2.3M iterat                                                                                      set allocating          time:   [2.8888 us 5.9331 us 9.6853 us]
                        change: [+53.967% +191.08% +338.05%] (p = 0.00 < 0.05)
                        Performance has regressed.
```

jmwill86:
I've just tried this on Mac with the normal terminal and with Kitty and can't replicate this issue. If the terminal window is small enough that that text moves to 2 lines then `print_overwritable` (used here as far as I can see) will replace the second line only.

The desired effect of `print_overwritable` is that it's a temporary message when running bench anyway, so it seems to me that this is by design rather than a bug that it can't be copy and pasted?

jyn514:
[...]

PR #564, opened 2022-03-07
addition of fn print_newline to stop same line behaviour when using c…

jmwill86:
…arriage return then resizing from smaller terminal windows and reset of last_line_len. Plus some formatting caused by rustfmt-preview

### DIFF
--- a/src/report.rs
+++ b/src/report.rs
@@ -1,20 +1,17 @@
-use crate::stats::univariate::outliers::tukey::LabeledSample;
-use crate::{csv_report::FileCsvReport, stats::bivariate::regression::Slope};
-use crate::{html::Html, stats::bivariate::Data};
-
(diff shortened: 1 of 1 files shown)


jmwill86:
Proposed fix for #435

jmwill86:
Just noticed this has been sitting here a while @bheisler? If there's anything I need to do then let me know :+1:

lemmih:
> Just noticed this has been sitting here a while @bheisler? If there's anything I need to do then let me know 👍

It might take a while before this will be reviewed. Could help if you include a summary of the problem (including why it affects some terminals more than others), how your code solves the problem, and if there will be any visible changes other than fixing #435.

bheisler:
Sorry, I haven't had much energy for open source lately.

I'm kind of tired today and perhaps missing something obvious, but could you explain the problem and the solution in more detail? I'm not sure I understand either what the problem is or why this would fix it.

Issue #582, opened 2022-06-13
Add documents and examples for async functions


daixi98:
When I used criterion.rs to benchmark Tonic RPC calls, I spent some time on configuring async functions. It would be better if there are examples of async functions in criterion.rs docs and books.

PR #583, opened 2022-06-13
Add docs and examples for async functions

daixi98:
Issue #582 : Add examples of async functions for criterion.rs docs.


bheisler:
Hello! I believe you'll need to add a dev-dependency on tokio in order to fix up the broken doctests.

PR #595, opened 2022-07-11
deps(criterion-plot): update 'cast' to 0.3, following the rest of criterion

poliorcetics:
Since there was no conflict at compilation before, `cast` should be an internal-only dependency of
`criterion-plot` which means a patch update is possible without breaking compatibility AFAICT.

### DIFF
--- a/plot/Cargo.toml
+++ b/plot/Cargo.toml
@@ -12,7 +12,7 @@ categories = ["visualization"]
 license = "MIT/Apache-2.0"
 
 [dependencies]
-cast = "0.2"
+cast = "0.3"
 itertools = "0.10"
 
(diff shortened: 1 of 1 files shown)


bheisler:
Hello! Thanks for the suggestion.

It is my policy not to accept simple version-number changes. However, this made me realize that I needed to publish a new version of `criterion-plot` as well, so I've updated all of its dependencies and published a new version. The current version of Criterion-rs still specifies a dependency on version 0.4.4 of `criterion-plot`. I vaguely recall that Cargo will accept versions that only differ in the last digit of the version number so it may just work, or you may have to wait until the next version of Criterion.rs.

In any case, thanks for the reminder!

PR #598, opened 2022-07-14
criterion-plot: update cast to 0.3

AlexTMjugador:
This major version of the `cast` crate matches the one used by the main `criterion` crate, eliminating the usage of several version of the aforementioned crate by Cargo, which very slightly reduces build times. As far as I know, there are no breaking changes that affect this crate.


AlexTMjugador:
Nevermind, I've just noticed that https://github.com/bheisler/criterion.rs/pull/595 beat me to it!

PR #605, opened 2022-07-26
Remove dependency on `itertools`

CosmicHorrorDev:
This PR removes the dependency on `itertools` in favor of slightly more verbose implementations. The two usages of itertools were for

## `izip!()`

Which was just swapped for chaining `.zip()` followed by a final `.map()` to pack the elements into a single tuple

## And `.group_by()`

Which was replaced with grouping everything into a `BTreeMap` first, then iterating over the `key -> group` pairs.

This implementation is slightly different since `.group_by()` will chunk the groups which means that the same key can appear multiple times if it is separated by another key. The implementation in this PR groups **all** values with the same key regardless of if they are separated by another key, so I'd like for someone else to double check that this still follows intended behavior


lemmih:
Anyone available to double check that this still works as expected?

briansmith:
I ended up submitting my own PR to do this, slightly differently. PTAL if you are interested: #876.

CosmicHorrorDev:
Super happy to have someone else pick up the torch. Thanks @briansmith !

PR #608, opened 2022-08-16
Fix user guides html report path

asze17:
From `target/criterion/report/index.html` to `target/criterion/reports/index.html`, which is the actual path the index.html is located in.


bheisler:
Thanks for the pull request!

PR #616, opened 2022-09-17
Update docs: Make fn fibonacci() public in code example

memark:
The current "Getting started" code example doesn't work. The `fn fibonacci()` must be public for the code to compile.


memark:
Any comments, @bheisler?

lemmih (approved):

PR #620, opened 2022-09-30
Update Latest Version in README.md

mj-nehme:


weihanglo:
I believe there are several other places need to update. See #622.

PR #621, opened 2022-10-06
Update html_report.md

### LINKED ISSUE #618: HTML Report Not Generated, Gnuplot Installed
I'm following the getting started guide and I am not seeing HTML reports being generated, in spite of `gnuplot` existing and being on `PATH`. I'm on Linux for what it's worth, amd64.

I'm on `criterion` `0.4.0` according to Cargo.lock, `criterion-plot` `0.5.0`. 

Relevant sections of `Cargo.toml`:

```toml
[dev-dependencies]
criterion = "0.4"

[[bench]]
name = "example"
harness = false
```

And my benchmark, at `benches/example.rs`:

```rust
use criterion::{black_box, criterion_group, criterion_main, Criterion};

fn fibonacci(n: u64) -> u64 {
    match n {
        0 => 0,
        1 => 1,
        n => fibonacci(n - 1) + fibonacci(n - 2),
    }
}

fn criterion_benchmark(c: &mut Criterion) {
    c.bench_function("fib 20", |b| b.iter(|| fibonacci(black_box(20))));
}

criterion_group!(benches, criterion_benchmark);
criterion_main!(benches);
```

As you can see, I'm following the getting started guide exactly. In fact, my repository is open source: [github.com/naftulikay/rust-fuzzy-datetime](https://github.com/naftulikay/rust-fuzzy-datetime/tree/0d75594846315db9623388009cb181e087ba9bbc).

Here's what's in `target/criterion`:

```text
target/criterion/
└── fib 20
    ├── base
    │   ├── benchmark.json
    │   ├── estimates.json
    │   ├── sample.json
    │   └── tukey.json
    └── new
        ├── benchmark.json
        ├── estimates.json
        ├── sample.json
        └── tukey.json

3 directories, 8 files
```

Proving `gnuplot` is there:

```shell
$ which gnuplot
/usr/bin/gnuplot
```

I'm running `cargo bench`, and I see the JSON but no HTML report being generated.

Is there a command line argument I need to be passing?

frederikhors:
Fixes https://github.com/bheisler/criterion.rs/issues/618.

### DIFF
--- a/book/src/user_guide/html_report.md
+++ b/book/src/user_guide/html_report.md
@@ -1,7 +1,10 @@
 # HTML Report
 
-### You should use feature "html_reports" for this.
-
+Starting with Criterion.rs 0.4.0 HTML reports must be explicitly enabled via the `html_reports` [feature](https://doc.rust-lang.org/cargo/reference/features.html#dependency-features):
+```toml
+[dev-dependencies]
+criterion = {version = "0.4.0", features = ["html_reports"] }
+```
 Criterion.rs can generate an HTML report displaying the results of the benchmark under
 `target/criterion/reports/index.html`. By default, the plots are generated using
 [gnuplot](http://www.gnuplot.info/) if it is available, or the
--- a/book/src/user_guide/html_report.md
+++ b/book/src/user_guide/html_report.md
@@ -1,5 +1,7 @@
 # HTML Report
 
+### You should use feature "html_reports" for this.
+
 Criterion.rs can generate an HTML report displaying the results of the benchmark under
 `target/criterion/reports/index.html`. By default, the plots are generated using
 [gnuplot](http://www.gnuplot.info/) if it is available, or the


theCollectiv:
It's still very confusing. The docs do not say, **how** to use html_report and in the current version "criterion 0.4.0" there is no way to get reporting or plotting working (by using criterion.rs only).

In version "criterion 0.3.6" you get at least a warning about the coming change, but in version "criterion 0.4.0" the change is implemented (somehow), but not clearly documented and how to deal with it.

See also: Unable to generate html-report (or plots) · Issue #632 · bheisler/criterion.rs -> https://github.com/bheisler/criterion.rs/issues/632

frederikhors:
What do you suggest?

theCollectiv:
I think, that the line 'You should use feature "html_reports" for this.' could provide more detailed information. So that a user, who is not familiar with criterion (or the changes in version history), can quickly understand, how to get reports/plotting working in version 0.4.0.

The doc for the `Command-Line Options`(a) is deprecated, because `cargo bench -- --plotting-backend gnuplot` or  `cargo bench -- --plotting-backend plotters` are disabled in version 0.4.0.

I'm not sure, how feature "html_reports" should be enabled (whether via commandline option or via Cargo.toml ...but both do not work in version 0.4.0).

So there are just 2 possible ways to get it working at the moment:
* switch back to version 0.3.6 (and then is everything working according to the current docs)
* use version 0.4.0, install cargo-criterion and use `cargo criteron` to get it working (reports will be generated automatically)

So I suggest
* keep the warning when using `cargo bench` in version 0.4.0
```
WARNING: HTML report generation will become a non-default optional feature in Criterion.rs 0.4.0.
This feature is being moved to cargo-criterion (https://github.com/bheisler/cargo-criterion) and will be 
optional in a future version of Criterion.rs. To silence this warning, either switch to cargo-criterion 
or enable the 'html_reports' feature in your Cargo.toml.
```
* mark the doc for the `Command-Line Options`(a), `cargo bench -- --plotting-backend gnuplot` and  `cargo bench -- --plotting-backend plotters` as deprecated in version 0.4.0
* 'You should use feature "html_reports" for this.' maybe better replaced by explicitly referring to 'install cargo-criterion and use `cargo criteron` to generate reports by default, if using version 0.4.0'.
* Update the docs referring to `html generation` (b) and `plotting` (c), that there is a breaking change to generate reports from version 0.3.6 to 0.4.0 
 
(a) [Command-Line Options](https://bheisler.github.io/criterion.rs/book/user_guidecommand_line_options.html#command-line-options)
(b) [HTML Report - Criterion.rs Documentation](https://bheisler.github.io/criterion.rs/book/user_guide/html_report.html)
(c) [Plots & Graphs - Criterion.rs Documentation](https://bheisler.github.io/criterion.rs/book/user_guide/plots_and_graphs.html)

book/src/user_guide/html_report.md
```diff
@@ -1,5 +1,7 @@
 # HTML Report
 
+### You should use feature "html_reports" for this.
+
```
Marcono1234: I agree with @theCollectiv, especially for users who have not worked with Cargo features much yet it would be good to explicitly show how to enable this. Also accessibility-wise this probably should not be a header (`###` prefix). What about the following?
````suggestion
Starting with Criterion.rs 0.4.0 HTML reports must be explicitly enabled via the `html_reports` [feature](https://doc.rust-lang.org/cargo/reference/features.html#dependency-features):
```toml
[dev-dependencies]
criterion = {version = "0.4.0", features = ["html_reports"] }
```
````

maxgitt: Having this PR landed would've saved me a ton of time because it wasn't clear to me in the docs why I wasn't seeing html_reports. https://stackoverflow.com/questions/75808243/rust-criterion-wont-create-report-html

Marcono1234:
The documentation has been improved by #662, however unfortunately `html_report.md` has not been adjusted. So the pull request here seems to still be relevant.

@lemmih, do you think it would be worth including the changes here as well?

maxgitt:
Per my comment above. Landing this PR landed would've saved me a ton of time. Please save the next person from having to do some digging.

PR #622, opened 2022-10-08
0.4 everywhere

weihanglo:
Just notice v0.4 got released! Thank you for creating such a useful benchmark framework!

I found that some guides still refer to 0.3, so update them all.


weihanglo:
Oops. #620 already addresses that. Closing. Sorry for not looking into existing PRs first.

PR #656, opened 2023-02-16
Add optional async setup for iter_batch

### LINKED ISSUE #576: `AsyncBencher::iter_batched()` setup closure to return a `Future`
Just like the Routine, the Setup is also likely to be async, when benching async code. Is there any particular reason why the setup is not async?

samtay:
This improves usability in async contexts. For example you might have

```rust
struct Struct;

impl Struct {
    pub fn new() -> Self { Self }
    pub async fn init_phase(&mut self) {}
    pub async fn something_worth_benchmarking(&self) {}
}
```

but currently you can't use `init_phase` in the `iter_batch` setup.

I decided to just offer a new variant so that this wouldn't be a breaking change. I've also got a different version (ref [7d12e73](https://github.com/samtay/criterion.rs/commit/7d12e7365a8d9f751c72b6a7f1e3cd263ccfd15b)) that updates all the async benchers to allow async setups. If that's preferable, let me know and I'll push that instead. 

Closes #576.

### DIFF
--- a/src/bencher.rs
+++ b/src/bencher.rs
@@ -7,6 +7,9 @@ use crate::measurement::{Measurement, WallTime};
 use crate::BatchSize;
 
 #[cfg(feature = "async")]
+use futures::{future, stream::FuturesUnordered, StreamExt};
+
+#[cfg(feature = "async")]
 use std::future::Future;
 
 #[cfg(feature = "async")]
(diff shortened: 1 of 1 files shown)


r-warfield:
This would be really helpful!! I would prefer not to use `iter_custom`.

Issue #658, opened 2023-02-22
crash in quick mode without "--noplot" when there is no time variation


KonradHoeffner:
When I benchmark in normal mode, all is fine but with -- --quick I get `assertion failed: slice.len() > 1 && slice.iter().all(|x| !x.is_nan())`:

```
cargo bench -- --save-baseline main
...
Benchmarking ??? (all)/0.2 all str triples: Warming up for 3.0000 s
Warning: Unable to complete 10 samples in 5.0s. You may wish to increase target time to 83.1s.
??? (all)/0.2 all str triples
                        time:   [9.8296 s 9.9077 s 9.9875 s]
                        change: [+3531649676% +3559731589% +3589878809%] (p = 0.01 < 0.05)
                        Performance has regressed.
```

```
cargo bench -- --quick --save-baseline main
??? (all)/0.2 all str triples
                        time:   [1.0000 ms 1.0000 ms 1.0000 ms]
                        change: [-99.990% -99.990% -99.990%] (p = 0.01 < 0.05)
                        Performance has improved.
thread 'main' panicked at 'assertion failed: slice.len() > 1 && slice.iter().all(|x| !x.is_nan())', /home/konrad/.cargo/registry/src/github.com-1ecc6299db9ec823/criterion-0.4.0/src/stats/univariate/sample.rs:31:9
stack backtrace:
   0: rust_begin_unwind
             at /rustc/246eae2fab54a5139365c4e1a08c5724fb385fbf/library/std/src/panicking.rs:579:5
   1: core::panicking::panic_fmt
             at /rustc/246eae2fab54a5139365c4e1a08c5724fb385fbf/library/core/src/panicking.rs:64:14
   2: core::panicking::panic
             at /rustc/246eae2fab54a5139365c4e1a08c5724fb385fbf/library/core/src/panicking.rs:114:5
   3: criterion::plot::gnuplot_backend::pdf::pdf_small
   4: <criterion::plot::gnuplot_backend::Gnuplot as criterion::plot::Plotter>::pdf
   5: <criterion::html::Html as criterion::report::Report>::measurement_complete
   6: criterion::analysis::common
   7: criterion::benchmark_group::BenchmarkGroup<M>::bench_function
   8: bench::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

Benchmark code:

```
let mut group = c.benchmark_group("??? (all)");
group.sample_size(10);  
[...]

PR #659, opened 2023-02-23
Fix bug where time over the maximum would be reported as 1ms in quick mode

KonradHoeffner:
See issue https://github.com/bheisler/criterion.rs/issues/658.
See also the minimal working example at <https://github.com/KonradHoeffner/criterion-quick-bug>.
This will report the correct time of the first and only iteration in quick mode.
Note that it will still crash with plotting due to the values being the same, so `cargo bench -- --quick --noplot` must be used.

### DIFF
--- a/src/routine.rs
+++ b/src/routine.rs
@@ -102,7 +102,6 @@ pub(crate) trait Routine<M: Measurement, T: ?Sized> {
 
             // Early exit for extremely long running benchmarks:
             if time_start.elapsed() > maximum_bench_duration {
-                let t_prev = 1_000_000f64;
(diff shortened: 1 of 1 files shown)


lemmih:
Sigh, might need to bump MSRV to 1.60

Victor-N-Suadicani:
Also just ran into this bug. Will this be released soon?

KonradHoeffner:
@Victor-N-Suadicani: The fix is included in the new version 0.5.0, however it will still crash in quick mode in those cases without the "--noplot" parameter.

PR #664, opened 2023-03-17
Mark black_box function as unsafe

kitcatier:
Hello, I found a soundness issue in this crate.
https://github.com/bheisler/criterion.rs/blob/2f5360737807cbe90d149db6199783236f0ef634/src/lib.rs#L165-L171
[https://doc.rust-lang.org/std/ptr/fn.read_volatile.html](url)
The unsafe function called needs to ensure that the parameter must be:

- src must be valid for reads.
- src must be properly aligned.
- src must point to a properly initialized value of type T.

and it is not a good choice to mark the entire function body as unsafe, which will make the caller ignore the safety requirements that the function parameters must guarantee, the developer who calls the black_box function may not notice this safety requirement.
Marking them unsafe also means that callers must make sure they know what they're doing.

### DIFF
--- a/src/bencher.rs
+++ b/src/bencher.rs
@@ -246,14 +246,14 @@ impl<'a, M: Measurement> Bencher<'a, M> {
 
         if batch_size == 1 {
             for _ in 0..self.iters {
-                let input = black_box(setup());
+                let input = unsafe { black_box(setup()) };
 
                 let start = self.measurement.start();
                 let output = routine(input);
                 let end = self.measurement.end(start);
                 self.value = self.measurement.add(&self.value, &end);
 
-                drop(black_box(output));
+                drop(unsafe { black_box(output) });
             }
         } else {
             let mut iteration_counter = 0;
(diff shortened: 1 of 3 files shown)


taiki-e:
> ```rs
> let ret = std::ptr::read_volatile(&dummy); 
> ```

This line reads the value not from `dummy`, but from a reference to `dummy` (`&dummy`). In other words, it copies `dummy` itself. And a reference satisfies all the safety requirements you mentioned.

I don’t think there is a soundness issue here except that we need to replace mem::forget with ManuallyDrop to avoid aliasing issues.

SUPERCILEX:
Can we just remove this function entirely? Std black_box was stabilized in 1.66.

h33p:
> Hello, I found a soundness issue in this crate.
> 
> https://github.com/bheisler/criterion.rs/blob/2f5360737807cbe90d149db6199783236f0ef634/src/lib.rs#L165-L171
> 
> [https://doc.rust-lang.org/std/ptr/fn.read_volatile.html](url)
> The unsafe function called needs to ensure that the parameter must be:
> 
>     * src must be valid for reads.
> 
>     * src must be properly aligned.
> 
>     * src must point to a properly initialized value of type T.
> 
> 
> and it is not a good choice to mark the entire function body as unsafe, which will make the caller ignore the safety requirements that the function parameters must guarantee, the developer who calls the black_box function may not notice this safety requirement. Marking them unsafe also means that callers must make sure they know what they're doing.

By passing `T` by value all of the required safety invariants are valid. There is no safety issue with this function. That said, removing `black_box` completely probably makes sense since it's been stabilized.

lemmih:
Closing. Let's switch to [std::hint::black_box](https://doc.rust-lang.org/std/hint/fn.black_box.html) once our MSRV hits 1.66

PR #679, opened 2023-04-27
Bump clap to version 4

sjackman:
Bump `clap` to version 4.
Bump MSRV to 1.64 for `clap`.

- Use `value_parser` for numeric value types
- Change `possible_values` to `value_parser`
- Change `hidden(true)` to `hide(true)`
- Change `value_of` and `value_of_t_or_exit` to `get_one`
- Change `is_present` to `get_flag` or `contains_id`
- Remove `takes_value(true)`, since `num_args(1)` is the default
- Use `num_args(0)` where needed

See https://github.com/clap-rs/clap/releases/tag/v3.2.0


src/lib.rs
```diff
@@ -920,50 +927,51 @@ https://bheisler.github.io/criterion.rs/book/faq.html
...
                 eprintln!("Warning: --noplot will be ignored when running with cargo-criterion. Use `cargo criterion --plotting-backend disabled -- <args>` instead.");
             }
-            if let Some(backend) = matches.value_of("plotting-backend") {
+            if let Some(backend) = matches.get_one::<String>("plotting-backend") {
                 eprintln!("Warning: --plotting-backend will be ignored when running with cargo-criterion. Use `cargo criterion --plotting-backend {} -- <args>` instead.", backend);
             }
-            if let Some(format) = matches.value_of("output-format") {
+            if let Some(format) = matches.get_one::<String>("output-format") {
                 if format != "criterion" {
                     eprintln!("Warning: --output-format will be ignored when running with cargo-criterion. Use `cargo criterion --output-format {} -- <args>` instead.", format);
```
sunshowers: Hmm, is there a reason for the different ways in which `get_one` is called here?

sunshowers (approved):
Looks pretty good to me! The changes are mostly mechanical. Just one question.

src/lib.rs
```diff
@@ -920,50 +927,51 @@ https://bheisler.github.io/criterion.rs/book/faq.html
             .get_matches();
 
         if self.connection.is_some() {
-            if let Some(color) = matches.value_of("color") {
+            if let Some(color) = matches.get_one("color").map(String::as_str) {
```
sjackman: @sunshowers This line was the odd one out.
I've changed
```rust
            if let Some(color) = matches.get_one("color").map(String::as_str) {
```
to
```rust
            if let Some(color) = matches.get_one::<String>("color") {
```

src/lib.rs
sjackman: When used in an `if` statement where the type needs to be comparable to a `&str`, it ought to be
```rust
if let Some(…) = matches.get_one::<String>("…") {
```

When used in an `if` statement where the type needs to be exactly a `&str`, it ought to be
```rust
match matches.get_one("…").map(String::as_str) {
````

I've fixed one case on line 930. Thanks for catching that style nit!

sjackman:
Thank you for merging!

sjackman:
I see that the last stable release was version 0.4.0 on 2022-09-10. No rush, but might there be a stable release on the horizon that includes this change?

sjackman:
@lemmih Thank you for the releases, David!

PR #680, opened 2023-05-13
Post v0.4 CHANGELOG updates

nbarrios1337:
Note: criterion would likely need a release-related PR bumping all the `criterion = "0.4"` instances across the documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Replaced atty dependency with is-terminal
 - MSRV bumped to 1.64
 - Upgraded clap dependency to v4
+- Upgraded tempfile dependency to v3.5.0
 
 ### Fixed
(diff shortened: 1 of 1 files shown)


sunshowers (commented):
Looks good to me! I also reviewed #679 which looks great.

nbarrios1337:
@lemmih Happy to set this to a ready state, but I think we need more details on the upcoming release to finalize the changelog update:

- Will the upcoming release require a minor or patch version bump?
- Should we have a similar PR addressing all the instances of "criterion = 0.4" in the examples and sample code?
- Should this PR merge before or after the release?

lemmih:
> @lemmih Happy to set this to a ready state, but I think we need more details on the upcoming release to finalize the changelog update:
> 
> * Will the upcoming release require a minor or patch version bump?

Semantic versioning is fuzzy regarding changes to MSRV but let's make it a major release out of an abundance of caution.

> * Should we have a similar PR addressing all the instances of "criterion = 0.4" in the examples and sample code?

Yes.

> * Should this PR merge before or after the release?

Before.

lemmih (approved):

PR #695, opened 2023-06-22
Depend on itertools >=0.10, <0.12

sjackman:
Depend on itertools >=0.10, <0.12

PR #699, opened 2023-07-10
ci: Remove AppVeyor

sjackman:
Remove AppVeyor.

On 2023-07-07 in <https://github.com/bheisler/criterion.rs/pull/695#issuecomment-1625830334> @lemmih wrote…
> You could make a separate PR that gets rid of AppVeyor.


lemmih:
Hmm, I don't have permission to remove the AppVeyor webhook. Think we have to keep the appveyor file for now but change it to not build anything.

@bheisler Mind removing the AppVeyor webhook?

bheisler:
Removed the webhook. Should I delete the old Travis-CI one as well?

lemmih:
> Removed the webhook. Should I delete the old Travis-CI one as well?

Yes. Thank you.

PR #701, opened 2023-07-11
Reexport std::hint::black_box

### LINKED ISSUE #700: The `black_box` function can be replaced with `std::hint::black_box`
The function `std::hint::black_box` is a stable wrapper around the same intrinsic that `test::black_box` uses, and could replace Criterion's own `black_box` function on all release channels. This function has been stable since 1.66, meaning that the last three minor versions of Rust are supported.

Spartan2909:
Fixes #700.


waywardmonkeys:
Would it make sense to possibly have a `black_box` function still, but mark it as deprecated and suggest calling `std::hint::black_box` instead, so that people can migrate?

Spartan2909:
That's also a good option. @bheisler any opinions?

lemmih:
> That's also a good option. @bheisler any opinions?

I think that sounds like a good idea.

bheisler:
Oh, wow. They finally stabilized `std::hint::black_box`? I didn't think that would ever happen.

Yeah, deprecating our hacky replacement for `black_box` seems good to me, but keep in mind the minimum-supported-Rust-version policy. It wouldn't seem right to deprecate in favor of a stdlib function that hasn't existed for at least three major versions.

waywardmonkeys:
Stable (const unstable) since 1.66 ... so that's 5 versions now if I count correctly. :)

Spartan2909:
I've added a deprecated wrapper for `std::hint::black_box`.

waywardmonkeys:
You will probably need to bump the MSRV in CI and elsewhere since it is currently stating and testing 1.64

waywardmonkeys:
Alternatively, deprecate in favor of std::hint but keep the old implementation for those that don’t want to update? Up to the maintainers!

d-e-s-o:
Any chance the `real_blackbox` feature could be cleaned up in the process? Or is this a compat hazard?

Spartan2909:
I've bumped the MSRV and removed the use of the `test` feature. @d-e-s-o removing a feature is a 
breaking change, but could be done with a minor version bump as Criterion is still <1.0. I'll leave this up to the maintainers.

lilyball:
I would really love to see this merged and a new release published. asan keeps sporadically triggering a false positive on the current `black_box` impl when used with zero-sized types so this change is more than just a performance fix.

Spartan2909:
@bheisler Is there anything blocking this?

waywardmonkeys:
@lemmih This can be closed now that #794 has landed. Thanks to @Spartan2909 for the work done here as helped pave the way!

PR #703, opened 2023-07-20
Enable `--help` messages and accept `--include-ignored` as no-op

### LINKED ISSUE #702: `--help` not supported?
I want to get a quick help of CLI parameters of criterion benches. Thus I tried:
`cargo bench --bench=mybench -- --help`
But it gives:
```
> cargo bench --bench=mybench -- --help
    Finished bench [optimized + debuginfo] target(s) in 0.03s
     Running benches/mybench.rs (/home/oxa/.cache/testt-target/release/deps/mybench-c6d0eae38261b75f)
Gnuplot not found, using plotters backend
error: unexpected argument found
error: bench failed, to rerun pass `--bench mybench`

Caused by:
  process didn't exit successfully: `/home/oxa/.cache/testt-target/release/deps/mybench-c6d0eae38261b75f --help --bench` (exit status: 2)
```

Even if I directly run the executable file, without cargo. It shows the same error.
```
> /home/oxa/.cache/testt-target/release/deps/mybench-c6d0eae38261b75f --help
Gnuplot not found, using plotters backend
error: unexpected argument found
```

I already have `harness = false` and other flags like `--baseline` all work. Just that `--help` is not recognized. Is this intentional or something going wrong?

I'm using criterion v0.5.1

oxalica:
- Since clap 4, `--help` requires the `help` feature to work. Fixes #702
- `--include-ignored` is a popular libtest parameter in CI to run all tests including non-ignored and ignored ones. In our case, we have no "ignored" semantics, thus it should be a no-op.


oxalica:
The CI error seems unrelated.

PR #707, opened 2023-08-09
Update CI

waywardmonkeys:


waywardmonkeys:
There's a chance that with the old action, it wasn't actually testing with the MSRV. I can reproduce this failure for 1.64 for `wasm32-wasi` locally. It also happens with 1.65 but not 1.66 or alter.

lemmih:
> There's a chance that with the old action, it wasn't actually testing with the MSRV. I can reproduce this failure for 1.64 for `wasm32-wasi` locally. It also happens with 1.65 but not 1.66 or alter.

Let's bump MSRV to 1.66?

waywardmonkeys:
I'll come back to this and dig more. It bothers me. This feature should be in Rust 1.63, so 1.64 shouldn't be an issue. It will be a couple of days though.

waywardmonkeys:
Looks like it is this: https://github.com/rust-lang/rust/issues/103306 which was fixed in 1.66 ... (it is 2:40am here and I should be asleep).

waywardmonkeys:
@lemmih  It seems a shame to update the MSRV so soon and for a fringe / experimental platform. Maybe it is okay to just skip this test except for on stable toolchain for now?

I've added a commit that does that in case you approve.

PR #712, opened 2023-08-21
`impl AsyncExecutor for tokio::runtime::Handle`

kpreid:
This will allow Tokio-based async benching without necessarily having access to the `Runtime` itself (e.g. inside of a `#[tokio::main]` function); a `Handle` is always obtainable via `Handle::current()`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 - gnuplot version is now correctly detected when using certain Windows binaries/configurations that used to fail
 
+### Added
+
+- Async benchmarking with Tokio may be done via a `tokio::runtime::Handle`, not only a `tokio::runtime::Runtime`
+
 ## [0.5.1] - 2023-05-26
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


lemmih:
Anyone up for reviewing this PR?

waywardmonkeys:
@kpreid Could you rebase this forward to fix the merge conflict?

@lemmih Since a `tokio::runtime::Handle` is just another way to get access to the currently running `Runtime`, I think this PR looks correct to me (without expert knowledge of tokio).

kpreid:
I've rebased, but I noticed something weird while working with git on it: the affected files seem to have CRLF line endings (even though the `.editorconfig` file suggests the repo should be all LF). This is probably just either harmless or something to fix in another PR, but I thought I'd mention it in case I'm missing something.

waywardmonkeys:
@kpreid Sorry to ask again, but could you rebase this (hopefully last time)?

kpreid:
Done. I notice that `async_executor.rs` now has LF endings, but `benchmarking_async.md` still has CRLF. The PR should be consistent with existing text in each file.

waywardmonkeys (approved):
@lemmih this should be good to go!

Thanks, @kpreid ! I will take a look at the line endings after this lands.

PR #713, opened 2023-08-24
Fix Label::is_outlier()

divergentdave:
This fixes an enum's predicate method that was incorrectly rewritten when adopting the `matches!` macro. Fixing this makes the "clean sample" blue dots show up on PDF charts again.


lemmih (approved):

waywardmonkeys:
@divergentdave Could you rebase this forward so that the CI can pass?

PR #722, opened 2023-09-12
Add support for reporting throughput and elements simultaneously.

vlovich:
Only hooked up for the CLI reporter for now, but it's pretty useful when doing data processing to understand both throughput variants when you're doing batch processing.

For the HTML report, this is equivalent to using Throughput::Elements although it may be useful to think about refining
it to visualize both.


vlovich:
@bheisler ping as requested in contribution guide

vlovich:
@waywardmonkeys are you maintaining this by any chance?

berkus:
Merged, can be closed.

PR #725, opened 2023-09-21
Increase criterion version in quick start

epompeii:
The version is currently set as `0.4` even though `0.5` has been available since May.


lemmih (approved):

waywardmonkeys:
@lemmih this was fixed in #759

PR #729, opened 2023-10-14
Clippy fixes

waywardmonkeys:

PR #731, opened 2023-10-14
ci: Fix MSRV build by precisely specifying some deps.

waywardmonkeys:


waywardmonkeys:
This also stops the build from cancelling all jobs once one fails, so that we can see the results for each version of Rust tested. Otherwise, the `stable` job fails (due to clippy) and we don't get results for the MSRV build.

waywardmonkeys:
#729 fixes the `stable` build and can be rebased forward and land once this one does.

lemmih:
How about joining the two PRs? I'd feel much better about merging a PR that passes all the checks.

waywardmonkeys:
@lemmih Okay, I've added the single commit here that should fix the current clippy build. The other PR contains further changes that go beyond just fixing the build.

PR #742, opened 2023-11-10
Book: CD to Github page

### LINKED ISSUE #677: Update the book and/or add a workflow for doing so
I noticed ´fibonacci´ not being public in the github pages book, but that was fixed a few months ago:
https://github.com/bheisler/criterion.rs/blob/57c4d847b730dee7ed04a8cf857d7b120f2ddabb/book/src/getting_started.md?plain=1#L9

If anyone wants to pick this up, adding a github actions workflow for this will save some time.
something like this: [example workflow](https://github.com/theseus-os/Theseus/blob/theseus_main/.github/workflows/docs.yaml)

kraktus:
close https://github.com/bheisler/criterion.rs/issues/677

see https://github.com/bheisler/criterion.rs/commit/d71f0908c82f44292d6c5a3a2f49b2ce378fa572 for technical details.

The current issue is that its posting the book at `https://bheisler.github.io/criterion.rs/` instead of `https://bheisler.github.io/criterion.rs/book/index.html` and I'm not sure how it'd plays with the master doc. before trying to fix this, I want to be sure the idea of CD for the book is validated.

In fact considering the "master" documentation still displays 0.3.4  and has not been updated for 2 years I think it'd be safe to scrape it and only serve the book: https://bheisler.github.io/criterion.rs/criterion/

### DIFF
--- a/.github/workflows/book.yaml
+++ b/.github/workflows/book.yaml
@@ -1,0 +1,62 @@
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - 'book/**'
+      - ".github/workflows/book.yaml"
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+# Sets the GITHUB_TOKEN permissions to allow deployment to GitHub Pages
+permissions:
+  contents: read
+  pages: write
+  id-token: write
+
+# Allow one concurrent deployment
+concurrency:
+  group: 'pages'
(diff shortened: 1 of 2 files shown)


kraktus:
gentle bump

berkus:
Merged, can be closed.

Documentation is published to https://criterion-rs.github.io, there is only book atm.

PR #744, opened 2023-11-22
Use alloca to address memory layout effects

PSeitz:
- replace collect with Vec::reserve
the iter collect pollutes the call stack (10 levels of calls)
- offset tests with alloca
As mentioned in #334, randomness in memory layout has a huge impact on the bench results. When I benchmark with criterion it's not uncommon to see effects of +-30%. Tests seem much more stable with alloca, there's still random 6% jumps in one test I've been running.

Adresses #334

### DIFF
--- a/src/routine.rs
+++ b/src/routine.rs
@@ -245,7 +245,7 @@ where
         let mut results = Vec::with_capacity(iters.len());
         results.resize(iters.len(), 0.0);
         for (i, iters) in iters.iter().enumerate() {
-            let stack_alloc = i % 2048;
+            let stack_alloc = i % 4096; // default page size
             #[cfg(any(target_family = "unix", target_family = "windows"))]
             {
                 alloca::with_alloca(
--- a/src/routine.rs
+++ b/src/routine.rs
@@ -245,10 +245,16 @@ where
         let mut results = Vec::with_capacity(iters.len());
         results.resize(iters.len(), 0.0);
         for (i, iters) in iters.iter().enumerate() {
-            b.iters = *iters;
-            (*f)(&mut b, black_box(parameter));
-            b.assert_iterated();
-            results[i] = m.to_f64(&b.value);
+            let stack_alloc = i % 2048;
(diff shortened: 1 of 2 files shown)


src/routine.rs
```diff
@@ -277,6 +293,8 @@ where
             }
 
             b.iters = b.iters.wrapping_mul(2);
+            b.iters = b.iters.min(64); // To make sure we offset the test at least with 0-64 bytes
+                                       // wit alloca
```
samueltardieu, line 297: Nit: typo ("wit" -> "with"). Also, the comment is typically placed above the line, not on the side.

src/routine.rs
```diff
@@ -242,15 +242,31 @@ where
...
-            .map(|iters| {
+        let mut results = Vec::with_capacity(iters.len());
+        results.resize(iters.len(), 0.0);
+        for (i, iters) in iters.iter().enumerate() {
+            let stack_alloc = i % 4096; // default page size
+            #[cfg(any(target_family = "unix", target_family = "windows"))]
+            {
+                alloca::with_alloca(
+                    stack_alloc, /* how much bytes we want to allocate */
+                    |_memory: &mut [core::mem::MaybeUninit<u8>] /* dynamically stack allocated slice itself */| {
```
samueltardieu, line 253: Maybe `stack_alloc` should be inlined here, to prevent getting a warning on platforms not selected by the `cfg` attribute. Using a name such as `_shifting_stack_space` or similar instead of `_memory` would be self-documenting.

samueltardieu (changes requested):
I like this changes. They should give better stability by increasing testing time variability.

FilipAndersson245:
Seems quite straightforward and if this helps in reducing random changes between runs I'm all for this.

waywardmonkeys:
@PSeitz Thanks for your contribution! Are you interested in updating this PR for the review comments or should we take it on?

berkus:
Merged, can be closed.

PR #759, opened 2024-02-22
Update criterion version in docs

FreezyLemon:
Noticed that the README still uses `0.4` in the Quickstart section. I updated all the examples that were using 0.4 in the docs. Not 100% sure about `html_report.md` because it explicitly mentions 0.4.0, but it's probably fine? Let me know

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ To start with Criterion.<span></span>rs, add the following to your `Cargo.toml` 
 
 ```toml
 [dev-dependencies]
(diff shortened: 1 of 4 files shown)


samueltardieu (approved):
LGTM

The change to `html_report.md` is fine also, the 0.4 which is mentioned is only the starting point for this feature.

waywardmonkeys (approved):
@lemmih I think this is good to go.

PR #760, opened 2024-02-26
Fix typos.

waywardmonkeys:


samueltardieu (approved):
CI failures are unrelated

waywardmonkeys:
@lemmih This updated now with an additional fix and since CI is passing on `master`, it is now passing here.

PR #767, opened 2024-03-24
Do not redundantly import `std::iter::IntoIterator`

samueltardieu:
Rust now warns about duplicate imports. `std::iter::IntoIterator` is part of the Rust prelude since Rust 2015. This prevents `criterion-plot` from compiling with recent Rust versions because of `#![deny(warnings)]`.

### DIFF
--- a/plot/src/axis.rs
+++ b/plot/src/axis.rs
@@ -1,7 +1,6 @@
 //! Coordinate axis
 
 use std::borrow::Cow;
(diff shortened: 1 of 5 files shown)


samueltardieu:
CI failure is unrelated (clap MSRV). It will be fixed by #768.

samueltardieu:
Merged and superseded by #768.

PR #775, opened 2024-04-25
Allow missing_docs in criterion_group

TomFryersMidsummer:
This lint gets flagged in user code if the missing_docs warning is enabled. Adding a doc comment to the function would also fix this.


berkus:
Obsoleted by https://github.com/criterion-rs/criterion.rs/pull/26

PR #793, opened 2024-07-25
deps: Require itertools 0.13

waywardmonkeys:
Itertools API changed slightly and they deprecated `group_by`,so just require latest for now.

Also, remove uses of `#[macro_use]` and `extern crate` with `itertools`.


lemmih (approved):

sjackman:
@lemmih Could you please push a new stable release to crates.io including this change when you have a chance? Thank you for your work on Criterion.rs, David!

djugei:
criterion-plot still uses v0.10

PR #798, opened 2024-07-27
Use dep syntax for enabling optional dependencies.

waywardmonkeys:
This prevents creating implicit features and also prevents these from showing up as features that can be enabled when using `cargo add criterion`.


waywardmonkeys:
This now also improves the configuration of `async_futures` to be more typical.

PR #802, opened 2024-07-27
Improve docs.

waywardmonkeys:
* More cross-linking.
* Remove `extern crate` stuff from doc tests / snippets.


waywardmonkeys:
@lemmih This one should be good to go too. (But I wrote it, so of course I think that.)

PR #768, opened 2024-03-24
Make CI green again

samueltardieu:
Version of clap above 4.4.9 require Rust 1.74, which exceeds criterion's MSRV of 1.70, and some entities belong to the Rust profile.

PR #803, opened 2024-08-07
Unpin `clap` dependency

jhpratt:
I'm unable to see why this was pinned in the first place. As pinning is unidiomatic and causes unexpected build failures, let's unpin it.


waywardmonkeys:
If you look at the history for that line, you'll see this was done in #768 because of MSRV issues.

If you want to unpin this, then it would be better to add `Cargo.lock` (remove it from `.gitignore` as well) and have `Cargo.lock` use versions that are compatible with our MSRV.

jhpratt:
I've done just that. Committing `Cargo.lock` is considered best practice in Rust nowadays regardless :slightly_smiling_face:

kaivol:
Are there any plans to merge this PR?

The current situation makes it impossible to use criterion in a cargo workspace where newer versions of `clap` are used.

lemmih:
Need someone to review this PR.

waywardmonkeys:
@lemmih This PR is fine as-is, and then we can do follow ups to potentially update the` Cargo.lock` some and perhaps evaluate what the current MSRV should be.

waywardmonkeys:
(I didn't look at the actual current contents of the `Cargo.lock` and assumed it was valid.)

waywardmonkeys (approved):

jhpratt:
I assure you that the `Cargo.lock` was auto-generated at the time of opening the PR. I have no doubt that there are things out of date in it by now, however. It's trivial to regenerate it from scratch if preferred.

lemmih:
Replaced by #858

PR #813, opened 2024-09-08
Add rust-version 1.70

jayvdb:
Current reason for msrv is clap, c.f. https://github.com/bheisler/criterion.rs/pull/803


waywardmonkeys (approved):
I think this is fine. Could probably apply the same to some of the other crates or in the future, switch to workspace inheritance to configure it once for all crates within the workspace.

PR #830, opened 2024-12-07
Add Becnher.iter_reuse

IgorErin:
Hello! In my scenario for benchmarking asynchronous runtime, I needed to reuse a preallocated buffer that moved through several closures and was cleared after each start. 

It seemed to me that this feature is not currently supported natively in criterion, so I suggest my own version, in the hope that it will be useful to others. 

Thank you in advance!


IgorErin:
@bheisler The guide says to pull you in case of a long wait.

PR #821, opened 2024-10-20
fix build on minimal versions

Toru3:
38b6d7fc71a6e9485e6cc8bef4179735ba9f4cf8
```terminal
$ cargo +nightly update -Z minimal-versions && cargo +stable --version && cargo +stable check
    Updating crates.io index
     Locking 0 packages to earliest compatible versions
note: pass `--verbose` to see 161 unchanged dependencies behind latest
cargo 1.82.0 (8f40fc59f 2024-08-21)
    Checking serde v1.0.85
    Checking regex v1.5.0
    Checking criterion-plot v0.5.0 (/home/tfukaya/codes/criterion.rs/plot)
    Checking clap v4.4.9
error[E0433]: failed to resolve: use of undeclared crate or module `syntax`
 --> /home/tfukaya/.cargo/registry/src/index.crates.io-6f17d22bba15001f/regex-1.5.0/src/literal/mod.rs:9:9
  |
9 |     use syntax::hir::literal::Literals;
  |         ^^^^^^ use of undeclared crate or module `syntax`

error[E0554]: `#![feature]` may not be used on the stable release channel
  --> /home/tfukaya/.cargo/registry/src/index.crates.io-6f17d22bba15001f/serde-1.0.85/src/lib.rs:85:35
   |
85 | #![cfg_attr(feature = "unstable", feature(specialization, never_type))]
   |                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error[E0554]: `#![feature]` may not be used on the stable release channel
  --> /home/tfukaya/.cargo/registry/src/index.crates.io-6f17d22bba15001f/serde-1.0.85/src/lib.rs:86:32
   |
86 | #![cfg_attr(feature = "alloc", feature(alloc))]
   |                                ^^^^^^^^^^^^^^

For more information about this error, try `rustc --explain E0433`.
error: could not compile `regex` (lib) due to 1 previous error
warning: build failed, waiting for other jobs to finish...
For more information about this error, try `rustc --explain E0554`.
error: could not compile `serde` (lib) due to 2 previous errors
```

892790297399fb8a35a34051e6bb14af6c1bedc2
```terminal
$ cargo +nightly update -Z minimal-versions && cargo +stable --version && cargo +stable check
    Updating crates.io index
     Locking 6 packages to earliest compatible versions
    Removing dtoa v0.4.0
    Updating itoa v0.3.0 -> v1.0.0 (available: v1.0.11)
[...]

PR #854, opened 2025-05-16
Add CI job checking library builds with oldest allowed dependencies

faern:
Prevents under-specifying our dependencies. Ensure the semver specs we write in our Cargo.toml are not allowing older versions than we actually build with. Catches issues such as the ones fixed in #821 before they are merged.

This type of check seems to have been present in the past, but removed at some point.

The current latest release `0.5.1` does not build with all dependencies downgraded to their oldest allowed versions. This is currently causing a bit of trouble for me. So I want to help by contributing a CI to check for this. I also hope a new release can be cut soon, since the issue has been fixed on `master` :pray:

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -10,6 +10,7 @@ on:
 name: tests
 env:
   CARGO_TERM_COLOR: always
+  MSRV: 1.80.0
 
 jobs:
   ci:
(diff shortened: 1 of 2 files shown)


lemmih:
I like it, but let's use MSRV or 1.87 instead of `+stable`.

faern:
Ok. I can change that. I usually use it with `+stable` and it usually don't cause problems. Because thanks to checking in the lockfile, the dependencies won't move. But now I changed it to MSRV. I also moved the MSRV definition to a variable, so it does not have to be updated in multiple places.

PR #875, opened 2025-07-10
Remove `itertools` dependency from the main crate.

briansmith:
`[T]::chunk_by` is stable as of Rust 1.77.


briansmith:
Folded into #876 to avoid merge conflicts between the two PRs.

PR #879, opened 2025-07-31
Extract shared package attributes into workspace

berkus:
- Synchronize version number across sub-crates.
- List workspace members explicitly.


berkus:
@lemmih here you go (for the next release the version should be bumped to 0.8 and criterion-plot will jump one version over, but this should be fine)

lemmih (approved):

PR #885, opened 2025-09-14
Set plot backgrounds to white for legibility on dark backgrounds

rben01:
This PR sets plot backgrounds to white in both the gnuplot and plotters backends. I'm not sure how to test this, so for verification, attached are the reports you get when running the following command with and without gnuplot installed:

```sh
cargo test --features plotters --features html_reports --test criterion_tests test_output_files
```

[criterion-output.zip](https://github.com/user-attachments/files/22321228/criterion-output.zip)

The reports are saved to a tempdir, so if you would like to replicate this yourself, you should place the following at the top of the test function and then add `-- --nocapture` to the command above.

```rust
let mut tempdir = temp_dir();
tempdir.disable_cleanup(true);
eprintln!("{tempdir:?}");  // copy this and cd to it in another shell
```

Screenshots of the reports in dark mode:

Gnuplot:
report

Plotters:
report

### DIFF
--- a/plot/src/lib.rs
+++ b/plot/src/lib.rs
@@ -478,6 +478,12 @@ impl Figure {
             }
         }
 
+        // fillstyle transparent solid 1 is required for noborder to take effect?
+        s.push_str(
+            "\nset object 1 rect from screen 0,0 to screen 1,1 behind \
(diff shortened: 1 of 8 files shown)


rben01:
I think the horizontal blue line in the second screenshot is an artifact of having had web inspector open and not inherent to the output.

Issue #826, opened 2024-11-11
Add support for documentation for macro invocations


MuhtasimTanmoy:
Problem
 
- Running `cargo clippy` with `rust.missing_docs = "warn"` lint enabled gives an error for the following macros.
  ```rust
  criterion_group!(benches, criterion_benchmark);
  criterion_main!(benches);
  ```

How to generate this issue?
-  [Clippy warning on documentation for macro invocations](https://stackoverflow.com/questions/79175703/clippy-warning-on-documentation-for-macro-invocations)

Possible solutions:
- [Generating documentation in macros](https://stackoverflow.com/questions/33999341/generating-documentation-in-macros)
- [Rust document macro invocations](https://amanjeev.com/blog/rust-document-macro-invocations)

Happy to open a PR if needed.

AliSajid:
I am giving this a thumbs up. I usually develop with `rust.missing_docs = "forbid"` but I had to go to `warn` for this so it would build. I'm also happy to contribute.

PR #888, opened 2025-10-01
Added `#[doc(hidden)]` to `criterion_group` to avoid raising `missing-docs` error

LucaCappelletti94:
As  per title, this PR simply ensures that the `criterion_group` macro no longer raises the `missing-docs` when such constraint is setup in the main cargo of a workspace. 

This PR closes issues such as #826


berkus:
This is fixed in #774 

Can be closed.

